PR #203, opened 2018-04-11
add auto detect board

Sneezry:


src/usbDetector.ts
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import * as os from 'os';
+import * as vscode from 'vscode';
+
+import {ConfigHandler} from './configHandler';
+import {ContentView} from './constants';
+
+const DEVICE_INFO = [{deviceId: 'devkit', vendorId: 0x0483, productId: 0x374b}];
```
dooriya: I think these magic numbers are already in constants.ts

PR #204, opened 2018-04-11
Auto download board package

Sneezry:


src/ArduinoPackageManager.ts
```diff
@@ -0,0 +1,51 @@
...
+        'https://raw.githubusercontent.com/VSChina/azureiotdevkit_tools/master/package_azureboard_index.json',
+    packageName: 'AZ3166',
+    architecture: 'stm32f4'
+  }
+};
+
+export class ArduinoPackageManager {
+  private static async setAdditionalUrl(url: string) {
+    const existedUrls = vscode.workspace.getConfiguration().get<string[]>(
+        'arduino.additionalUrls');
```
dooriya: use constant since it is used in several places

Sneezry: We need design constant better. Currently constant is not in good structure. Let's make it in a standalone task.

PR #223, opened 2018-04-23
extend the structure to support more boards 

lirenhe:
TODO:  update all examples to add 

"IoTWorkbench.BoardId": "devkit",


src/ArduinoPackageManager.ts
```diff
@@ -50,15 +41,24 @@ export class ArduinoPackageManager {
...
 
-  static async installBoard(boardId: string) {
-    const board = BOARD_INFO[boardId];
-    if (!board) {
+  static async installBoard(context: vscode.ExtensionContext, boardId: string) {
+    const boardList = context.asAbsolutePath(
+        path.join(FileNames.resourceFolderName, FileNames.boardListFileName));
+    const boardsJson = require(boardList);
+
+    const board = boardsJson.boards.filter((template: Board) => {
```
Sneezry: use `find` here. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

src/ArduinoPackageManager.ts
```diff
@@ -50,15 +41,24 @@ export class ArduinoPackageManager {
...
+      return template.id === boardId;
+    });
+
+    if (!board || !board[0]) {
       return;
     }
```
Sneezry: check if installation is undefined here, and return if the board doesn't have installation property.

src/Models/AZ3166Device.ts
```diff
@@ -74,6 +74,7 @@ export class AZ3166Device implements Device {
   private deviceFolder: string;
   private extensionContext: vscode.ExtensionContext;
   private sketchName = '';
+  static boardId = 'devkit';
```
Sneezry: should it be private?

Sneezry: let's create a getter to prevent changing boardId outside

src/Models/IoTProject.ts
```diff
@@ -234,10 +240,12 @@ export class IoTProject {
...
     workspace.folders.push({path: constants.deviceDefaultFolderName});
 
-    const device = new AZ3166Device(
-        this.extensionContext, deviceDir, projectTemplateItem.sketch);
-    this.componentList.push(device);
-
+    if (boardId === AZ3166Device.boardId) {
+      const device = new AZ3166Device(
+          this.extensionContext, deviceDir, projectTemplateItem.sketch);
+      workspace.settings[`IoTWorkbench.${ConfigKey.boardId}`] = boardId;
```
Sneezry: we should have a config helper to handle config settings

src/projectInitializer.ts
```diff
@@ -188,7 +187,8 @@ export class ProjectInitializer {
             }
 
             const project = new IoTProject(context, channel);
-            return await project.create(rootPath, result[0], openInNewWindow);
+            return await project.create(
+                rootPath, result[0], boardSelection.id, openInNewWindow);
```
Sneezry: let use `find` to handle the result previous as well

PR #231, opened 2018-04-25
add config

Sneezry:


src/Models/IoTButtonDevice.ts
```diff
@@ -163,6 +169,189 @@ export class IoTButtonDevice implements Device {
...
+    const uri = constants.accessPointHost + constants.wifiPath;
+
+    const res = await this.setConfig(uri, data);
+
+    return res;
+  }
+
+  async configHub() {
+    let deviceConnectionString =
+        ConfigHandler.get<string>(ConfigKey.iotHubDeviceConnectionString);
```
lirenhe: better check whether the connection string exists

src/Models/IoTButtonDevice.ts
```diff
@@ -163,6 +169,189 @@ export class IoTButtonDevice implements Device {
...
+      uri,
+      method: 'POST',
+      timeout: constants.timeout,
+      headers: constants.requestHead,
+      json
+    };
+
+    const res = await request(option);
+
+    return res;
```
lirenhe: how can we tell user whether the operation succeed or failed?

PR #318, opened 2018-06-27
Add raspberry pi

Sneezry:


resources/raspberrypi/simple_iothub.js
```diff
@@ -0,0 +1,33 @@
+const clientFromConnectionString = require('azure-iot-device-mqtt').clientFromConnectionString;
+const Message = require('azure-iot-device').Message;
```
dooriya: Keep consistency on the const value naming styling

Sneezry: `clientFromConnectionString` is a function, and `Message` is a class. Neither of them is variable. So the naming style should okay here.

PR #328, opened 2018-07-09
Fix the confusing steps in example documents

ArthurMa1978:


docs/iot-devkit/devkit-shakeshake.md
```diff
@@ -22,8 +22,8 @@ An active Azure subscription. If you do not have one, you can register via one o
 
 ### Start VS Code
 
-- Start VS Code.
-- Connect the DevKit to your computer.
+- Make sure your DevKit is not connected to your PC.
```
Sneezry: Now the example can popup even DevKit is connected before start vscode

PR #427, opened 2018-09-15
Add the arduino package path for linux platforms

flipbug:
The arduino package path is currently only defined for windows and mac os, which prevents the extension from being fully usable while developing on a linux platform. The fix is to simply add the required path for linux.


Sneezry:
Thanks for your contribution!

PR #431, opened 2018-09-18
Support multiple devices in the workspace

ab-sin-the:


ab-sin-the:
Find some bugs when using devkit

PR #469, opened 2018-11-05
Fix error message and update azure function

lirenhe:


src/Models/AzureFunctions.ts
```diff
@@ -283,7 +288,7 @@ export class AzureFunctions implements Component, Provisionable, Deployable {
             'azureFunctions.deploy', azureFunctionsPath, functionAppId);
       } else {
         const subPath =
-            path.join(azureFunctionsPath, 'bin/Release/netstandard2.0/publish');
+            path.join(azureFunctionsPath, 'bin/Release/netcoreapp2.1/publish');
```
Sneezry: Is there a way to load the version dynamically instead of hard coding?

Sneezry: Or we have to update code every time when Azure Functions updates?

lirenhe: As i know, they did not provide any config for this.

PR #543, opened 2018-12-20
remove functions from hard dependency

Sneezry:


src/Models/AzureFunctions.ts
```diff
@@ -185,6 +208,11 @@ export class AzureFunctions implements Component, Provisionable, Deployable {
   }
 
   async provision(): Promise<boolean> {
+    const isFunctionsExtensionAvailable = await AzureFunctions.isAvailable();
```
lirenhe: Is it possible to check whether the project contains azure function and then ask user to install it?

Sneezry: Yeah. `AzureFunctions.isAvailable` is a static method, so we can call it anywhere without context.

I thinks this is more like a user experience topic. Should we show the notification as soon as the user opens a project contains Functions component or only when user executes provision command?

src/Models/AzureFunctions.ts
```diff
@@ -85,6 +85,33 @@ export class AzureFunctions implements Component, Provisionable, Deployable {
     return this.componentType;
   }
 
+  static async isAvailable(): Promise<boolean> {
+    if (!vscode.extensions.getExtension(
+            'ms-azuretools.vscode-azurefunctions')) {
```
lirenhe: Move it to constants

src/Models/AzureFunctions.ts
```diff
@@ -85,6 +85,33 @@ export class AzureFunctions implements Component, Provisionable, Deployable {
     return this.componentType;
   }
 
+  static async isAvailable(): Promise<boolean> {
+    if (!vscode.extensions.getExtension(
+            'ms-azuretools.vscode-azurefunctions')) {
+      const choice = await vscode.window.showInformationMessage(
+          'Azure Functions extension has been disabled or not installed. Please enable Azure Functions extension or install it from marketplace, and try again.',
```
lirenhe: Is it better to use:    Azure Functions extension is required for the current project. Do you want to install it from marketplace?

src/Models/AzureFunctions.ts
```diff
@@ -85,6 +85,33 @@ export class AzureFunctions implements Component, Provisionable, Deployable {
...
+    if (!vscode.extensions.getExtension(
+            'ms-azuretools.vscode-azurefunctions')) {
+      const choice = await vscode.window.showInformationMessage(
+          'Azure Functions extension has been disabled or not installed. Please enable Azure Functions extension or install it from marketplace, and try again.',
+          'Install Azure Functions Extension', 'Close');
+      if (choice === 'Install Azure Functions Extension') {
+        vscode.commands.executeCommand(
+            'vscode.open',
+            vscode.Uri.parse(
+                'vscode:extension/ms-azuretools.vscode-azurefunctions'));
```
lirenhe: ms-azuretools.vscode-azurefunctions should be a constant value

src/Models/IoTProject.ts
```diff
@@ -143,6 +143,10 @@ export class IoTProject {
         this.componentList.push(functionApp);
       }
 
+      this.componentList.forEach(item => {
+        item.checkPrerequisites();
```
lirenhe: checkPrerequisites is async function

Sneezry: The order is not necessary here, so `await` is not required.

PR #593, opened 2019-03-19
Emphasize step for Instance of IoT Device Provisioning Service

DanKline:
Emphasize step for Instance of IoT Device Provisioning Service so it is not easily missed for users with Azure subscriptions


Sneezry:
Hi @DanKline thanks for your PR. `.vs` folder needs to be removed.

DanKline:
I'm seeing this error on the build engine which is outside of the scope of my changes to the md file:

Downloading html-pipeline-2.9.0 revealed dependencies not in the API or the
lockfile (nokogiri (>= 1.4)).
Either installing with `--full-index` or running `bundle update html-pipeline`
should fix the problem.
Command exited with code 34

Sneezry:
You can try to execute these commands to remove the committed `.vs` folder:

```bash
git rm --cached .vs
git commit -m "clean up"
git push
```

PR #627, opened 2019-05-30
Provison Azure Service

dilin-MS:
1. Bug fix: Enable return to last step when discovering device
2. Scaffold common files when load an iot project
3. Add "Provison Azure Service"
4. Create DevKit project with Azure Function
5. Archieve Dockerfiles on ACR

### DIFF
--- a/resources/Arduino/devkit/template.json
+++ b/resources/Arduino/devkit/template.json
@@ -12,6 +12,32 @@
                     "targetPath" : "./device"
                 }
             ]
+        },
+        {
+            "description": "",
+            "detail": "Create project that connects to Azure IoT Hub and processes device data further in Azure Functions.",
+            "label": "With Azure Functions",
+            "type": "AzureFunctions",
+            "templateFilesInfo": [
+                {
+                    "fileName": "simple_iothub.ino",
+                    "sourcePath" : ".",
+                    "targetPath" : "./device"
+                }
+            ]
+        },
+        {
+            "description": "",
+            "detail": "Create project that processes device data from IoT Hub in Stream Analytics and store the result in Cosmos DB.",
+            "label": "With Stream Analytics and Cosmos DB",
+            "type": "StreamAnalytics",
+            "templateFilesInfo": [
+                {
+                    "fileName": "simple_iothub.ino",
+                    "sourcePath" : ".",
+                    "targetPath" : "./device"
+                }
+            ]
         }
     ]
 }
--- a/resources/Arduino/devkit/template.json
+++ b/resources/Arduino/devkit/template.json
@@ -25,19 +25,6 @@
                     "targetPath" : "./device"
                 }
             ]
-        },
-        {
-            "description": "",
-            "detail": "Create project that processes device data from IoT Hub in Stream Analytics and store the result in Cosmos DB.",
-            "label": "With Stream Analytics and Cosmos DB",
-            "type": "StreamAnalytics",
-            "templateFilesInfo": [
-                {
-                    "fileName": "simple_iothub.ino",
-                    "sourcePath" : ".",
-                    "targetPath" : "./device"
-                }
-            ]
         }
     ]
 }
--- a/resources/Linux/raspberrypi/IoTHub/devcontainer.json
+++ b/resources/Linux/raspberrypi/IoTHub/devcontainer.json
@@ -1,0 +1,10 @@
+{
+  "image": "rpi_app",
+  "name": "rpi_app_container",
+  "dockerFile": "Dockerfile",
+  "extensions": [
+    "ms-vscode.cpptools",
+    "twxs.cmake",
+    "vsciot-vscode.vscode-iot-workbench"
+  ]
+}
--- a/src/FileUtility.ts
+++ b/src/FileUtility.ts
@@ -3,112 +3,97 @@
 import * as fs from 'fs-plus';
 import * as path from 'path';
 import * as sdk from 'vscode-iot-device-cube-sdk';
+import {ScaffoldType} from './constants';
 
 export class FileUtility {
 
-  // File-related functions to help create files on workspace path
-  static async existsInWorkspace(localPath: string): Promise<boolean> {
-    return new Promise((resolve: (exist: boolean) => void) => {
-      fs.stat(localPath, (error: Error | null) => {
-        if (error) {
-          resolve(false);
+  static async exists(type: ScaffoldType, path: string): Promise<boolean> {
+    if (type === ScaffoldType.local) {
+      return await sdk.FileSystem.exists(path);
+    } else {
+      return new Promise((resolve: (exist: boolean) => void) => {
+        fs.stat(path, (error: Error | null) => {
+          if (error) {
+            resolve(false);
+            return;
+          }
+          resolve(true);
           return;
-        }
-        resolve(true);
-        return;
+        });
       });
-    });
-  }
-
-  static async writeFileInWorkspace(filePath: string, data: string | Buffer): Promise<void> {
-    return new Promise(async (resolve: (value?: void) => void, reject) => {
-      await fs.writeFile(filePath, data, (err) => {
-        if (err) {
-          reject(err);
-        }
-        return;
-      });
-      resolve();
-    });
-  }
-
-  static async directoryExistsInWorkspace(dirPath: string): Promise<boolean> {
-    try {
-      return fs.statSync(dirPath).isDirectory();
-    } catch (e) {
-      return false;
     }
   }
 
-  static async mkdirRecursivelyInWorkspace(dirPath: string): Promise<void> {
-    if (await FileUtility.directoryExistsInWorkspace(dirPath)) {
+  static async writeFile(type: ScaffoldType, filePath: string, data: string | Buffer): Promise<void> {
+    if (type === ScaffoldType.local) {
+      return await sdk.FileSystem.writeFile(filePath, data);
+    } else {
+      return new Promise(async (resolve: (value?: void) => void, reject) => {
+        await fs.writeFile(filePath, data, (err) => {
+          if (err) {
+            reject(err);
+          }
+          return;
+        });
+        resolve();
+      });
+    }
+    
+  }
+
+  static async directoryExists(type: ScaffoldType, dirPath: string): Promise<boolean> {
+    if (type === ScaffoldType.local) {
+      if (!await sdk.FileSystem.exists(dirPath)) {
+        return false;
+      }
+      const isDirectory = await sdk.FileSystem.isDirectory(dirPath);
+      return isDirectory;
+    } else {
+      try {
+        return fs.statSync(dirPath).isDirectory();
+      } catch (e) {
+        return false;
+      }
+    }
+  }
+
+  static async mkdirRecursively(type: ScaffoldType, dirPath: string): Promise<void> {
+    if (await FileUtility.directoryExists(type, dirPath)) {
       return;
     }
     const dirname = path.dirname(dirPath);
     if (path.normalize(dirname) === path.normalize(dirPath)) {
       fs.mkdirSync(dirPath);
-    } else if (await FileUtility.directoryExistsInWorkspace(dirname)) {
+    } else if (await FileUtility.directoryExists(type, dirname)) {
       fs.mkdirSync(dirPath);
     } else {
-      await FileUtility.mkdirRecursivelyInWorkspace(dirname);
+      await FileUtility.mkdirRecursively(type, dirname);
       fs.mkdirSync(dirPath);
     }
   }
 
-  static async fileExistsInWorkspace(filePath: string): Promise<boolean> {
-    try {
-      return fs.statSync(filePath).isFile();
-    } catch (e) {
-      return false;
-    }
-  }
-
-  static async readFileInWorkspace(filePath: string, encoding: string): Promise<string> {
-    return fs.readFileSync(filePath, encoding);
-  }
-
-  // File-related functions to help create files on a local path
-  static async existsInLocal(localPath: string): Promise<boolean> {
-    return await sdk.FileSystem.exists(localPath);
-  }
-
-  static async writeFileInLocal(filePath: string, data: string | Buffer): Promise<void> {
-    return await sdk.FileSystem.writeFile(filePath, data);
-  }
-
-  static async directoryExistsInLocal(dirPath: string): Promise<boolean> {
-    if (!await sdk.FileSystem.exists(dirPath)) {
-      return false;
-    }
-    const isDirectory = await sdk.FileSystem.isDirectory(dirPath);
-    return isDirectory;
-  }
-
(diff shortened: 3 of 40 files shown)


src/FileUtility.ts
```diff
@@ -0,0 +1,114 @@
+'use strict';
+
+import * as fs from 'fs-plus';
+import * as path from 'path';
+import * as sdk from 'vscode-iot-device-cube-sdk';
+
+export class FileUtility {
+
+  // File-related functions to help create files on workspace path
+  static async existsInWorkspace(localPath: string): Promise<boolean> {
```
lirenhe: How is this path related to workspace?

lirenhe: I think the api signature should be changed to
fileExists(path, location).    If location is local, use the API from fs-plus, else use the API from cube.

src/Models/AZ3166Device.ts
```diff
@@ -206,8 +206,9 @@ export class AZ3166Device extends ArduinoDeviceBase {
       throw new Error('Unable to find the board in the config file.');
     }
 
-    // await ScaffoldGenerator.scaffolIoTProjectdFiles(this.projectFolder, this.vscodeFolderPath, 
-    //   this.boardFolderPath, this.devcontainerFolderPath, this.board.id);
+    const scaffoldGenerator = new ScaffoldGenerator();
+    await scaffoldGenerator.scaffolIoTProjectdFilesInWorkspace(this.projectFolder, this.vscodeFolderPath,
```
lirenhe: scaffolIoTProjectdFilesInWorkspace  typo

what is the return value of this function?

src/Models/AzureComponentConfig.ts
```diff
@@ -40,38 +40,60 @@ export class AzureConfigFileHandler {
...
   constructor(projectRoot: string) {
     this.projectRootPath = projectRoot;
     this.configFilePath = path.join(
         this.projectRootPath, AzureComponentsStorage.folderName,
         AzureComponentsStorage.fileName);
   }
 
-  createIfNotExists() {
+
+  async createIfNotExistsInWorkspace() {
```
lirenhe: combine those 2 API, use a variable to indicate whether this is local/remote in constructor.

src/Models/AzureComponentConfig.ts
```diff
@@ -107,11 +129,10 @@ export class AzureConfigFileHandler {
     }
   }
 
-  getComponentIndexById(id: string) {
+  async getComponentIndexById(id: string) {
     try {
-      const azureConfigs =
-          JSON.parse(fs.readFileSync(this.configFilePath, 'utf8')) as
-          AzureConfigs;
+      const azureConfigContent = await FileUtility.readFileInLocal(this.configFilePath, 'utf8');
```
lirenhe: no idea why use async here. any benefit?

src/Models/AzureComponentConfig.ts
```diff
@@ -120,11 +141,10 @@ export class AzureConfigFileHandler {
     }
   }
 
-  getComponentById(id: string) {
+  async getComponentById(id: string) {
```
lirenhe: same above. No need to be async

resources/Arduino/devkit/template.json
```diff
@@ -12,6 +12,19 @@
...
+            "detail": "Create project that connects to Azure IoT Hub and processes device data further in Azure Functions.",
+            "label": "With Azure Functions",
+            "type": "AzureFunctions",
+            "templateFilesInfo": [
+                {
+                    "fileName": "simple_iothub.ino",
+                    "sourcePath" : ".",
+                    "targetPath" : "./device"
+                }
+            ]
```
lirenhe: For the support of Azure Functions, there are some additional tool to be installed on target machine.
https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azurefunctions
Does your docker file contain all those information?

dilin-MS: The current version of iot workbench will suggest install the azure-function extension when user choose "With Azure Functions" devkit project. This method encounters one issue that currently vscode-insider does not provide accessible link like` vscode:extension/ms-azuretools.vscode-azurefunctions`.

As you suggested, we can directly choose a docker image with azure function extension pre-installed for users if they choose to create a "With Azure Functions" devkit project.
I will fix the issue next working day.

src/FileUtility.ts
dilin-MS: * **Workspace path**  looks like: `/workspace/myproject/devkit/sample.ino`
* **Local path** looks like: `C:/users/username/Documents/IoTWorkbenchProjects/projects/myproject/devkit/sample.ino`

I use `async` to align two sets of API(for local path and workspace path) which will save much duplicate code. And the local file-related functions is said will be refactored to be `async`. Did I misunderstand that?

Also, your suggestion to pass the path as an argument may be a better implementation. I will check whether this method can work well in the specific scenario.

src/Models/AzureComponentConfig.ts
dilin-MS: The reason is that API we used from fs-plus is **sync**, but API from cube is **async**. If we combine the two API into one, we have to use **async** since we are not certain which API will be used.

resources/Arduino/devkit/template.json
dilin-MS: Fixed.

resources/Linux/raspberrypi/IoTHub/devcontainer.json
```diff
@@ -0,0 +1,10 @@
+{
+  "image": "rpi_app",
+  "name": "rpi_app_container",
+  "dockerFile": "Dockerfile",
+  "extensions": [
+    "ms-vscode.cpptools",
+    "twxs.cmake",
```
lirenhe: Why is this extension needed?

dilin-MS: Because we have CMakeLists.txt as one of our sample files. This extension is for intelliSense of CMake file.

src/FileUtility.ts
```diff
@@ -0,0 +1,107 @@
+'use strict';
+
+import * as fs from 'fs-plus';
+import * as path from 'path';
+import * as sdk from 'vscode-iot-device-cube-sdk';
+import {ScaffoldType} from './constants';
+
+export class FileUtility {
+
+  static async exists(type: ScaffoldType, path: string): Promise<boolean> {
```
lirenhe: better call it
  static async fileExists(environmentType: EnvironmentType, path: string): Promise<boolean> {

dilin-MS: This functions can judge not only **file** but also **directory**. So we use `exists` as the function name.
We can change the `type` argument to `environmentType`.

src/FileUtility.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (type === ScaffoldType.local) {
+      return await sdk.FileSystem.exists(path);
+    } else {
+      return new Promise((resolve: (exist: boolean) => void) => {
```
lirenhe: I think you could still use fs.statsync in this case to avoid the callback.
You also need to check the return value to detect whether this is a file or a dir.

lirenhe: I found there is a directoryExists api below, what does this API plan to detect?

src/FileUtility.ts
dilin-MS: But actually, the type is not indicating the environment whether is local or in workspace, since scenario can be that when we are in remote side and want to create a new project. In this situation we will pass a `local` type as argument in order to use the cube api to scaffold files in local filesystem.
i.e. the `type` argument is more about where to scaffold the new files, in local filesystem or in workspace.
Will it be better if we use the `scaffoldType` as the argument name?

src/constants.ts
```diff
@@ -57,6 +58,11 @@ export enum OperationType {
   upload = 'Upload device code'
 }
 
+export enum ScaffoldType {
```
lirenhe: use EnvironmentType with value: local, container, wsl, remotemachine

src/Models/AzureComponentConfig.ts
```diff
@@ -47,31 +47,29 @@ export class AzureConfigFileHandler {
         AzureComponentsStorage.fileName);
   }
 
-  createIfNotExists() {
+  async createIfNotExists(type: ScaffoldType) {
     const azureConfigs: AzureConfigs = {componentConfigs: []};
     const azureConfigFolderPath =
         path.join(this.projectRootPath, AzureComponentsStorage.folderName);
-    if (!fs.existsSync(azureConfigFolderPath)) {
+    if (!await FileUtility.exists(type, azureConfigFolderPath)) {
```
lirenhe: use dir exists?

src/Models/AzureComponentConfig.ts
```diff
@@ -107,11 +105,10 @@ export class AzureConfigFileHandler {
     }
   }
 
-  getComponentIndexById(id: string) {
+  async getComponentIndexById(id: string) {
     try {
-      const azureConfigs =
-          JSON.parse(fs.readFileSync(this.configFilePath, 'utf8')) as
-          AzureConfigs;
+      const azureConfigContent = await FileUtility.readFile(ScaffoldType.local, this.configFilePath, 'utf8');
```
lirenhe: why using scaffoldtype.local here and workspace in async getSortedComponents

src/FileUtility.ts
dilin-MS: The `exists` function can be removed. Fixed.

src/Models/AzureComponentConfig.ts
dilin-MS: Remove the `exists` api and fix this issue.

src/FileUtility.ts
```diff
@@ -0,0 +1,107 @@
...
+  static async directoryExists(type: ScaffoldType, dirPath: string): Promise<boolean> {
+    if (type === ScaffoldType.local) {
+      if (!await sdk.FileSystem.exists(dirPath)) {
+        return false;
+      }
+      const isDirectory = await sdk.FileSystem.isDirectory(dirPath);
+      return isDirectory;
+    } else {
+      try {
+        return fs.statSync(dirPath).isDirectory();
```
Sneezry: Do not use sync method

src/FileUtility.ts
```diff
@@ -0,0 +1,107 @@
...
+        return false;
+      }
+    }
+  }
+
+  static async mkdir(type: ScaffoldType, dirPath: string): Promise<void> {
+    if (type === ScaffoldType.local) {
+      await sdk.FileSystem.mkDir(dirPath);
+    } else {
+      fs.mkdirSync(dirPath);
```
Sneezry: Do not use sync method

src/FileUtility.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+
+  static async mkdirRecursively(type: ScaffoldType, dirPath: string): Promise<void> {
+    if (await FileUtility.directoryExists(type, dirPath)) {
+      return;
+    }
+    const dirname = path.dirname(dirPath);
+    if (path.normalize(dirname) === path.normalize(dirPath)) {
```
Sneezry: When will `dirPath` equal to its parent folder path?

src/FileUtility.ts
```diff
@@ -0,0 +1,107 @@
...
+    } else {
+      await FileUtility.mkdirRecursively(type, dirname);
+      await FileUtility.mkdir(type, dirPath);
+    }
+  }
+
+  static async fileExists(type: ScaffoldType, filePath: string): Promise<boolean> {
+    if (type === ScaffoldType.local) {
+      const directoryExists = await sdk.FileSystem.exists(filePath);
+      if (!directoryExists) {5
```
Sneezry: 5?

src/FileUtility.ts
```diff
@@ -0,0 +1,107 @@
...
+        return false;
+      }
+      const isFile = await sdk.FileSystem.isFile(filePath);
+      return isFile;
+    } else {
+      try {
+        return fs.statSync(filePath).isFile();
```
Sneezry: Do not use sync method

src/FileUtility.ts
```diff
@@ -0,0 +1,107 @@
...
+      } catch (e) {
+        return false;
+      }
+    }
+  }
+
+  static async readFile(type: ScaffoldType, filePath: string, encoding: string): Promise<string> {
```
Sneezry: The general read file interface should not fix the content type to string. If you only read file contains string, better handle at high level.

src/constants.ts
```diff
@@ -57,6 +58,11 @@ export enum OperationType {
...
+  local = 'local',
```
Sneezry: Better use uppercase

src/constants.ts
dilin-MS: The type is not simply about the environment information.

src/Models/AzureComponentConfig.ts
dilin-MS: Same as above.

src/FileUtility.ts
dilin-MS: Will fix.

src/FileUtility.ts
dilin-MS: Will fix.

src/FileUtility.ts
dilin-MS: dirPath = `a/b/c/.`
dirname = `a/b/c`

src/FileUtility.ts
dilin-MS: Typo. Will remove.

src/FileUtility.ts
dilin-MS: Fix.

src/FileUtility.ts
dilin-MS: Fixed.

src/constants.ts
dilin-MS: Fixed.

src/Models/ScaffoldGenerator.ts
```diff
@@ -177,54 +75,48 @@ export class ScaffoldGenerator {
...
-      const dockerfileSourcePath = path.join(
-        boardFolderPath, boardId, FileNames.dockerfileName);
-      const dockerfileContent = fs.readFileSync(dockerfileSourcePath, 'utf8');
-      await sdk.FileSystem.writeFile(dockerfileTargetPath, dockerfileContent);
-    } catch (error) {
-      throw new Error(`Create ${FileNames.dockerfileName} failed: ${error.message}`);
+      try {
+        const dockerfileSourcePath = path.join(
+          templateFolderPath, FileNames.dockerfileName);
+        const dockerfileContent = fs.readFileSync(dockerfileSourcePath, 'utf8');
```
Sneezry: Do not use sync method

dilin-MS: The whole refrator will be implemented in other pr.

PR #633, opened 2019-06-10
Fixing the "Open serial monitor" link

joe-bethke:
The previous link was not linking to the correct header, and was instead just linking to the top of the get started page. This minor fix links to the correct header.


lirenhe:
Thanks for your contribution @joe-bethke.

PR #638, opened 2019-06-18
Add isInternal property to identify internal users

blackchoey:


src/telemetry.ts
```diff
@@ -102,6 +103,9 @@ export async function callWithTelemetry(
     }
   }
 
+  properties['isInternal'] =
+      InternalConfig.isInternal === true ? 'true' : 'false';
```
lirenhe: why not using 

properties['isInternal'] = InternalConfig.isInternal  ?

blackchoey: If isInternal is not boolean type, the result will always be true. Just want to make the comparison more robust.

PR #647, opened 2019-06-26
Init change for Templates handling

lirenhe:


src/Models/Interfaces/Platfrom.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
dilin-MS: filename typo: Platform.ts

resources/templates/templates.json
```diff
@@ -0,0 +1,14 @@
+{
+    "templates": [
+        {
+            "platform": "Embedded Linux",
+            "name": "RaspBerry Pi with iothub",
```
dilin-MS: "Raspberry Pi"

PR #648, opened 2019-06-26
Rpi flow

dilin-MS:


resources/templates/platformlist.json
```diff
@@ -11,8 +11,8 @@
             "description": "Project based on Linux(Yocto/Ubuntu/Debian/...) Platform."
         },
         {
-            "name": "$(issue-opened) My device is not in the list...",
-            "id": "no_device",
+            "name": "$(issue-opened) My wanted platform is not in the list...",
```
lirenhe: let's use device for now

src/Models/IoTWorkbenchProjectBase.ts
```diff
@@ -1,12 +1,14 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+import {RetryOperation} from 'azure-iot-common';
```
lirenhe: do we need this?

src/Models/IoTWorkbenchProjectBase.ts
```diff
@@ -270,4 +274,21 @@ export abstract class IoTWorkbenchProjectBase {
   }
 
   abstract async handleLoadFailure(): Promise<boolean>;
+
+  static async generateIotWorkbenchProjectFile(
```
lirenhe: no need to be static

src/Models/IoTWorkbenchProjectBase.ts
dilin-MS: No.

resources/templates/platformlist.json
dilin-MS: Fixed.

src/Models/IoTWorkbenchProjectBase.ts
dilin-MS: In order to be called in `RaspberryPiDevice.ts`:  
`await IoTWorkbenchProjectBase.generateIotWorkbenchProjectFile(...);`

PR #649, opened 2019-06-27
support workspace mode for esp32

lirenhe:


src/Models/AZ3166Device.ts
```diff
@@ -147,11 +143,22 @@ export class AZ3166Device extends ArduinoDeviceBase {
       throw new Error('Unable to find the board in the config file.');
     }
 
+    const plat = os.platform();
+
     this.generateCommonFiles();
```
dilin-MS: Consider using the `IoTWorkbenchProjectBase.generateIotWorkbenchProjectFile()`
Notice the different path for Workspace / Container project's `.iotworkbenchprojectfile`.

src/Models/ArduinoDeviceBase.ts
```diff
@@ -194,7 +189,7 @@ export abstract class ArduinoDeviceBase implements Device {
   }
 
   async generateSketchFile(
-      sketchTemplateFileName: string, board: Board, boardInfo: string,
+      fileInfo: TemplateFileInfo, board: Board, boardInfo: string,
```
dilin-MS: Modify this function using FileUtility function instead of fs-plus file operation function. Or the creation of file will fail when create file from remote side.

dilin-MS: Can see the `containerization` branch for detail.

PR #735, opened 2019-08-28
configure devkit in container

RongXiefdu:


src/Models/AZ3166Device.ts
```diff
@@ -738,6 +744,48 @@ export class AZ3166Device extends ArduinoDeviceBase {
         });
   }
 
+  private async chooseCOM(comList: Array<any>): Promise<string> {
```
Sneezry: Cube SDK should provide `com` interface

PR #743, opened 2019-09-06
rename DPS terms in generated code for PnP

dooriya:
Align with PnP team, in all UIs and user visible surfaces, in order to keep the naming consistency, we are going to
- Change ‘Scope ID’ to ‘**ID Scope**’
- Change ‘registrationID’ to ‘**deviceID**’

@JerryYangKai once you've finished the rename items as follows, please let me know, I'll update the rest of documents/code snippet.
1. You need to update the [http server](https://github.com/microsoft/devkit-sdk/blob/master/AZ3166/src/cores/arduino/httpserver/app_httpd.cpp) to read DPS value from config portal and rename the DPS connection string format, and release a new package for AZ3166.

    - Current DPS connection string format: DPSEndpoint=[DPS global endpoint];ScopeId=[Scope ID];RegistrationId=[Registration ID];SymmetricKey=[symmetric key]
    - Expected format: DPSEndpoint=[DPS global endpoint];IdScope=[ID Scope];DeviceId=[Device ID];SymmetricKey=[symmetric key]

2. You need to refresh the DPS example.

### DIFF
--- a/resources/templates/ansic_cmake_iotcsaskey/Readme.md
+++ b/resources/templates/ansic_cmake_iotcsaskey/Readme.md
@@ -29,14 +29,14 @@ For more details about setting up your development environment for compiling the
     * Open `main.c`. Specify paramaters requested by the commented **TODO's** for your configuration.
 
       ```c
-      // TODO: Specify DPS scope ID if you intend on using IoT Central.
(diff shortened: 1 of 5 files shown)


dooriya:
Abandon this PR since it is duplicated with another PR

PR #749, opened 2019-09-14
Move hard-code DPS credentials out of main.c

dooriya:


xinyiz1226:
A general question is, do we want to have two ways for credentials? or passing via parameters is good enough. Need Liya's input for this.

dooriya:
> A general question is, do we want to have two ways for credentials? or passing via parameters is good enough. Need Liya's input for this.

> Resolved. As we synced offline, I've removed the option to config DPS credentials in source code.

PR #750, opened 2019-09-17
Upgrade gts to latest version and fix gts in source code

dooriya:


dooriya:
Abandon this change: due to the backward compatibility of gts.

PR #776, opened 2019-10-15
Serialport

JerryYangKai:


src/Models/AZ3166Device.ts
```diff
@@ -314,6 +321,61 @@ export class AZ3166Device extends ArduinoDeviceBase {
...
+        }
+      };
+
+      const deviceConnectionString = await vscode.window.showInputBox(option);
+
+      if (!deviceConnectionString) {
+        return false;
+      }
+
+      console.log(deviceConnectionString);
```
dooriya: I think we don't need to print the connection string here for security consideration.

PR #779, opened 2019-10-18
Fix ubuntu c_cpp_properties.json error

dilin-MS:


dilin-MS:
Re-open PR in https://github.com/microsoft/vscode-iot-workbench/pull/785.

PR #781, opened 2019-10-21
Update CodeGen flow to support various project types

dooriya:
## Changes:

* Refine the project template selection, available CodeGen project template: 
  - CMake Project on Windows
  - CMake Project on Linux
  - MXChip IoT DevKit Project

* Add new step to select Device SDK reference type:
   - Source Code
   - Vcpkg

* Update CodeGen config file according to the changed flow

## The New CodeGen flow should be:

* Select DCM file
* Create New Project or Regen for an existing project

* Select language
    
image

* Select IoT Connection type

image

* Select project template:

image

* Select Device C SDK reference type:

image


resources/templates/ansic_cmake_connectionstring/CMakeLists-Windows-SourceCode.txt
```diff
@@ -0,0 +1,46 @@
+#Copyright (c) Microsoft. All rights reserved.
+#Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#this is CMakeLists.txt for {PROJECT_NAME}
```
lirenhe: Not quite sure why adding '-Windows-SourceCode' after CMakeLists. Is this a valid cmake file?

dooriya: It is just a template  CMakeList file name, which is used to identified for different platform and different sdk refernce type, it will use the standard "CMakeLists.txt" file name when persistent on disc.

---
In reply to: [336844826](https://github.com/microsoft/vscode-iot-workbench/pull/781#discussion_r336844826) [](ancestors = 336844826)

PR #782, opened 2019-10-21
change binary file executability in server

dilin-MS:


src/Models/RaspberryPiDevice.ts
```diff
@@ -42,6 +42,32 @@ export class RaspberryPiDevice extends ContainerDeviceBase {
...
+      return;
+    }
+    const getBinaryFileNameCmd = `cat ${
+        cmakeFilePath} | grep 'set(binary_name' | cut -d ' ' -f2 | sed -e 's/^"//' -e 's/"$//' | tr -d '\n'`;
+
+    const binaryName = await execute(getBinaryFileNameCmd);
+    return binaryName;
+  }
+
+  private async EnableBinaryExecutability(
```
dooriya: Naming Consistency: first letter should be lower case

dilin-MS: Fixed.

src/Models/RaspberryPiDevice.ts
```diff
@@ -42,6 +42,32 @@ export class RaspberryPiDevice extends ContainerDeviceBase {
         DeviceType.Raspberry_Pi, templateFilesInfo);
   }
 
+  private async getBinaryFileName(): Promise<string|undefined> {
+    // Parse binary name from CMakeLists.txt file
+    const cmakeFilePath = path.join(this.projectFolder, 'CMakeLists.txt');
```
dooriya: There is a constant definition for “CMakeLists.txt” which you can leverage.

dilin-MS: The constant is in `DigitalTwinConstants.ts`. It seems not good to use constant in that scope.

src/Models/RaspberryPiDevice.ts
```diff
@@ -42,6 +42,32 @@ export class RaspberryPiDevice extends ContainerDeviceBase {
...
+    }
+    const getBinaryFileNameCmd = `cat ${
+        cmakeFilePath} | grep 'set(binary_name' | cut -d ' ' -f2 | sed -e 's/^"//' -e 's/"$//' | tr -d '\n'`;
+
+    const binaryName = await execute(getBinaryFileNameCmd);
+    return binaryName;
+  }
+
+  private async enableBinaryExecutability(
+      ssh: sdk.SSH, binaryName: string|undefined) {
```
dooriya: Per the method name, Suggest ensuring binaryName is a string and remove undefined here.

If a undefined binaryName is a ignorable case, please check in the caller, otherwise throw error for this case.

dilin-MS: Currently we change binary file executability when we can find the binary file in the correct place. Since we expose quite large extensible freedom to user, we will just ignore if we cannot find the binary file in the supposed place.
I will move the logic outside of function `enableBinaryExecutability` and add comment to describe the situation.

src/utils.ts
```diff
@@ -191,6 +191,20 @@ export async function selectWorkspaceItem(
                                  (await showOpenDialog(options))[0].fsPath;
 }
 
+export function execute(command: string): Promise<string> {
```
dooriya: **nit**: executeCmd will be more meaningful.

dilin-MS: Will fix.

src/Models/RaspberryPiDevice.ts
dooriya: Then move it to a generic place.

PR #783, opened 2019-10-21
Support linux cpp properties

dilin-MS:


resources/templates/esp32/c_cpp_properties_linux.json
```diff
@@ -0,0 +1,22 @@
+{
+    "configurations": [
+        {
+            "name": "Linux",
+            "defines": [
+                "ARDUINO=10800"
+            ],
+            "includePath": [
+                "${workspaceFolder}",
+                "{ROOTPATH}/.arduino15/packages/AZ3166/tools/**",
```
dooriya: Why we need to include "az3166" libraries in ESP32 project?

src/Models/ArduinoDeviceBase.ts
```diff
@@ -191,9 +192,23 @@ export abstract class ArduinoDeviceBase implements Device {
...
+                board.id, constants.cppPropertiesFileNameLinux));
+        const propertiesContentLinux =
+            fs.readFileSync(propertiesFilePathLinux).toString();
+        const rootPathPattern = /{ROOTPATH}/g;
+        const versionPattern = /{VERSION}/g;
+        const homeDir = await IoTWorkbenchSettings.getOs();
+        const replaceStr =
+            propertiesContentLinux.replace(rootPathPattern, homeDir)
+                .replace(versionPattern, this.version);
+        await FileUtility.writeFile(type, cppPropertiesFilePath, replaceStr);
```
dooriya: Similar code block as "win32" case, suggest moving to a wrapper method to improve readability.

src/Models/ArduinoDeviceBase.ts
```diff
@@ -191,9 +192,23 @@ export abstract class ArduinoDeviceBase implements Device {
...
       }
-      // TODO: Let's use the same file for Linux and MacOS for now. Need to
-      // revisit this part.
+      // TODO: Let's use the MacOS template file for OS system that is not win32
+      // / linux. Revisit this part if want to support other OS system.
```
dooriya: >OS system [](start = 61, length = 9)

Just OS, no more system

resources/templates/esp32/c_cpp_properties_linux.json
dilin-MS: It is a bug. Fixed.

src/Models/ArduinoDeviceBase.ts

src/Models/ArduinoDeviceBase.ts
dilin-MS: Not all the same since win32 will need changing `\`  to `\\`.

PR #786, opened 2019-10-23
[Bug fix] Configure container project

dilin-MS:
1. [Create c_cpp_properties.json file when configure container environment](https://dev.azure.com/mseng/VSIoT/_workitems/edit/1621581)
2. [Fix typo in build script](https://dev.azure.com/mseng/VSIoT/_workitems/edit/1621584)
3. Code Refactor: Reuse code to generate c_cpp_properties.json file for different OS


src/Models/ArduinoDeviceBase.ts
```diff
@@ -160,20 +160,46 @@ export abstract class ArduinoDeviceBase implements Device {
   abstract get version(): string;
 
   private async writeCppPropertiesFile(
-      rootPath: string, boardId: string, type: ScaffoldType,
-      cppPropertiesTemplateFileName: string,
-      cppPropertiesFilePath: string): Promise<void> {
+      boardId: string, type: ScaffoldType, plat: string): Promise<void> {
```
eriolchan: It is better to use full name as platform for easy to understand

src/Models/ArduinoDeviceBase.ts
```diff
@@ -160,20 +160,46 @@ export abstract class ArduinoDeviceBase implements Device {
...
+    }
+
     const cppPropertiesTemplateFilePath =
         this.extensionContext.asAbsolutePath(path.join(
             FileNames.resourcesFolderName, FileNames.templatesFolderName,
             boardId, cppPropertiesTemplateFileName));
     const propertiesContent =
-        fs.readFileSync(cppPropertiesTemplateFilePath).toString();
+        await FileUtility.readFile(type, cppPropertiesTemplateFilePath);
+    const propertiesContentString = propertiesContent.toString();
```
eriolchan: I have seen many code logic as this to replace version value in file in the past few reviews, can we make it a common function if we frequently use it?

dilin-MS: Yes. This is **the** common function.

dilin-MS: Yes. This is the common function.

dilin-MS: > I have seen many code logic as this to replace version value in file in the past few reviews, can we make it a common function if we frequently use it?

Yes. This is the common function.

src/Models/ArduinoDeviceBase.ts
dilin-MS: Will fix.

PR #788, opened 2019-10-25
Check cmake project before configured as Embeded Linux project

dilin-MS:


dooriya:
>     const indentationSpace = 4;

Normalize json intent

---
Refers to: src/Models/IoTContainerizedProject.ts:302 in 4b407d3. [](commit_id = 4b407d36f9f72fe56604577245d72528abb9e895, deletion_comment = False)

src/Models/IoTContainerizedProject.ts
```diff
@@ -258,19 +259,32 @@ export class IoTContainerizedProject extends IoTWorkbenchProjectBase {
 
   /**
    * Check if it is an external project.
-   * If external project, construct as RaspberryPi Device based container iot
+   * If external project, configure as RaspberryPi Device based container iot
    * workbench project.
    */
-  async constructExternalProjectToIotProject(scaffoldType: ScaffoldType) {
+  async ConfigExternalProjectToIotProject(scaffoldType: ScaffoldType):
```
eriolchan: Better to use camalCase for method name

src/ProjectEnvironmentConfiger.ts
```diff
@@ -122,8 +122,10 @@ export class ProjectEnvironmentConfiger {
 
       // If external project, construct as RaspberryPi Device based
       // container iot workbench project
-      await project.constructExternalProjectToIotProject(scaffoldType);
-
+      const res = await project.ConfigExternalProjectToIotProject(scaffoldType);
```
eriolchan: can merge to one statement like
```
if (!await project.ConfigExternalProjectToIotProject(scaffoldType)) {
  return false;
}
```

src/Models/IoTContainerizedProject.ts

src/ProjectEnvironmentConfiger.ts
dilin-MS: Fixed.

PR #813, opened 2019-11-11
Change extensionId in code if RC extension

dilin-MS:


scripts/modifyConfigIfRCVersion.js
```diff
@@ -2,9 +2,10 @@ const fs = require('fs');
 
 if (process.env.TRAVIS_TAG) {
   const ISTESTVERSION = /^v?[0-9]+\.[0-9]+\.[0-9]+-[rR][cC]$/.test(process.env.TRAVIS_TAG || '');
-  const packageJson = JSON.parse(fs.readFileSync('package.json'));
   if (ISTESTVERSION) {
```
eriolchan: I think camalCase would be more easy to read.

scripts/modifyConfigIfRCVersion.js
```diff
@@ -13,10 +14,25 @@ if (process.env.TRAVIS_TAG) {
...
 
-    // Delete
     delete packageJson.icon;
     delete packageJson.aiKey;
 
     fs.writeFileSync('package.json', JSON.stringify(packageJson, null, 2) + '\n');
+
+
+    // 2. Modify extensionId in files
+    const extensionIdPattern = /vsciot-vscode.vscode-iot-workbench/g;
```
eriolchan: Can we use dedicated file for test environment instead of dynamic replacement? It would be more maintainable and less error prone.

scripts/modifyConfigIfRCVersion.js
dilin-MS: fixed.

scripts/modifyConfigIfRCVersion.js
dilin-MS: Establish new task to track this issue. https://dev.azure.com/mseng/VSIoT/_workitems/edit/1644786

PR #819, opened 2019-11-14
clear PnP command and integrate with PnP extension

eriolchan:
- clear PnP related command palette
- clear PnP web view and template file
- integrate with PnP extension for code generation


src/DigitalTwin/DigitalTwinUtility.ts
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * check if digital twin extension is available,
+   * and init if available
+   * @param channel output channel
+   */
+  public static isAvailable(channel: vscode.OutputChannel): boolean {
+    const digitalTwins = getExtension(ExtensionName.DigitalTwins);
+    if (!digitalTwins) {
+      utils.channelShowAndAppendLine(
+        channel, 'Azure Digital Twins is not installed. Please install it from Marketplace.');
```
dooriya: >Azure Digital Twins  [](start = 18, length = 20)

=> Azure Digital Twins VS Code extension is not installed ...

src/DigitalTwin/DigitalTwinUtility.ts
```diff
@@ -0,0 +1,70 @@
...
+import { getExtension } from '../Models/Apis';
+import { ExtensionName } from '../Models/Interfaces/Api';
+import { DigitalTwinConstants } from './DigitalTwinConstants';
+
+/**
+ * Digital Twin extension utility
+ */
+export class DigitalTwinUtility {
+  // tslint:disable-next-line: no-any
+  private static instance: any;
```
dooriya: >private static instance: any; [](start = 2, length = 29)

An implication to use the APIs from DigitalTwinUtility is required to use DigitalTwinUtility.isAvailable() to check at first.
Can we initialize the "instance" by default rather than initialize it in the isAvailable() method?

The DigitalTwinUtility is more like a extension class for the DigitalTwins extension. how about using a more readable name for "instance" here?

src/DigitalTwin/DigitalTwinUtility.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * download dependent interface,
+   * return true if all interface is successfully downloaded, otherwise return false
+   * @param folder folder to download interface
+   * @param capabilityModelFile capability model file path
+   */
+  public static async downloadDependentInterface(folder: string, capabilityModelFile: string): Promise<boolean> {
+    try {
+      await DigitalTwinUtility.instance.downloadDependentInterface(folder, capabilityModelFile);
```
dooriya: >DigitalTwinUtility.instance [](start = 12, length = 27)

Null reference error will occur if caller forget to invoke the isAvailable() to initialize the 'instance' menber first.

PR #823, opened 2019-11-15
Fix configure device settings

dilin-MS:


vendor/node-usb-native/README.md
```diff
@@ -1,4 +1,4 @@
-This is a native package contains two modules: detector and serialport, detector provides the functionality of detecting usb changes and serialport provides the functionality of listing serial ports, openning serial ports and sending/receiving message to/from serial ports.
+This is a natvie package contains two modules: detector and serialport, detector provides the functionality of detecting usb changes and serialport provides the functionality of listing serial ports, openning serial ports and sending/receiving message to/from serial ports.
```
dooriya: typo：natvie => native

vendor/node-usb-native/lib/serialport.js
```diff
@@ -226,6 +226,20 @@ SerialPort.prototype.write = function(buffer, callback) {
...
   }
 
+  switch (ending) {
+    case 'Newline':
+      buffer = Buffer.concat([buffer, Buffer.from('\n')]);
+      break;
+    case 'Carriage return':
+      buffer = Buffer.concat([buffer, Buffer.from('\r')]);
+      break;
+    case 'Both NL & CR':
```
dooriya: You’ve already defined a constant value for ‘Both NL & CR’, suggest using Enum with string value here since there’re 3 cases here.

vendor/node-usb-native/src/detection.cpp
```diff
@@ -210,7 +210,7 @@ void StopMonitoring(const Nan::FunctionCallbackInfo<v8::Value>& args) {
 }
```
dooriya: Currently we keep two copies of the node-usb-native bits in both Arduino and Workbench extension. In future, we should consolidate this to reduce the maintain cost.

vendor/node-usb-native/src/combined.cpp
```diff
@@ -9,11 +9,11 @@
 #include <nan.h>
 
 extern "C" {
-	void init_serialport(v8::Handle<v8::Object> target);
+	void init_serialport( v8::Local<v8::Object> target);
```
dooriya: Extra space included here, I think this is due to copied from the Arduino repo, please also fix it here.

vendor/node-usb-native/README.md
dilin-MS: Fixed.

vendor/node-usb-native/lib/serialport.js
dilin-MS: Fixed.

vendor/node-usb-native/src/detection.cpp
dilin-MS: Set up task to track this issue.https://dev.azure.com/mseng/VSIoT/_workitems/edit/1647683

vendor/node-usb-native/src/combined.cpp
dilin-MS: Fixed.

vendor/node-usb-native/lib/serialport.js
```diff
@@ -23,6 +23,11 @@ var STOPBITS = [1, 1.5, 2];
 var PARITY = ['none', 'even', 'mark', 'odd', 'space'];
 var FLOWCONTROLS = ['xon', 'xoff', 'xany', 'rtscts'];
 var SET_OPTIONS = ['brk', 'cts', 'dtr', 'dts', 'rts'];
+const endings = {
+  Newline: 'Newline',
+  CarriageReturn: 'Carriage return',
+  BothNLAndCR: 'Both NL & CR'
+} ;
```
dooriya: Why not using Enum with string value

dilin-MS: JS does not have enum class.

PR #830, opened 2019-11-20
Only create project folder when generate code to file system

dooriya:


dooriya:
Change will be overwrite by PR #832. So abandon this one.

PR #852, opened 2019-11-28
Fix open in container bug

dilin-MS:
https://dev.azure.com/mseng/VSIoT/_workitems/edit/1648537


src/Models/IoTContainerizedProject.ts
```diff
@@ -147,10 +147,21 @@ export class IoTContainerizedProject extends IoTWorkbenchProjectBase {
...
     await this.openProject(this.projectRootPath, openInNewWindow);
   }
 
+  async openFolderInContainer(folderPath: string) {
+    if (!await FileUtility.directoryExists(ScaffoldType.Local, folderPath)) {
+      throw new Error(
+          `Fail to open folder in container. ${folderPath} does not exist.`);
+    }
+
+    setTimeout(
```
adashen: why need setTimeout here?

src/Models/IoTContainerizedProject.ts
```diff
@@ -147,10 +147,21 @@ export class IoTContainerizedProject extends IoTWorkbenchProjectBase {
...
+        () => vscode.commands.executeCommand(
+            'remote-containers.openFolder', vscode.Uri.file(folderPath)),
```
eriolchan: Since Cube extension is weakly dependent, what if remote extension is not installed here?

resources/templates/arm7/devcontainer.json
```diff
@@ -1,6 +1,5 @@
 {
-  "image": "armv7",
-  "name": "armv7_container",
+  "name": "Azure IoT Device C SDK for armv7 Linux",
```
adashen: Any reason now we need remove image name?

src/Models/IoTContainerizedProject.ts
```diff
@@ -147,10 +147,21 @@ export class IoTContainerizedProject extends IoTWorkbenchProjectBase {
           `Internal Error. Could not find iot workbench project file.`);
     }
 
-    // Open project
     await this.openProject(this.projectRootPath, openInNewWindow);
   }
 
+  async openFolderInContainer(folderPath: string) {
+    if (!await FileUtility.directoryExists(ScaffoldType.Local, folderPath)) {
```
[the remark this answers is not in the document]
dilin-MS: When trigger "open in container", we  are in local environment.

src/Models/IoTContainerizedProject.ts
dilin-MS: Wait for the telemetry to be sent before VS Code restart.

resources/templates/arm7/devcontainer.json
dilin-MS: VS Code update the `devcontainer.json` rule. In our scenario, we provide `dockerfile` and `devcontainer.json`. We only need to provide the `name` and `dockerfile` key for `remote-containers` extension to build a container for us. No need for `image` now.
See [devcontainer reference](https://code.visualstudio.com/docs/remote/containers#_devcontainerjson-reference).
`image` is only required when using an image instead of dockerfile.

src/Models/IoTContainerizedProject.ts
adashen: Better to use await telemetry send then trigger this command?

dilin-MS: Currently TelemetryRepoter does not provide async `sendTelemetryEvent` function. I have opened an issue on their repo.

src/Models/IoTContainerizedProject.ts

src/Models/IoTContainerizedProject.ts
adashen: Just check the repo. One question here, is telemetry.sendTelemetyrEvent really an async method? Or it is not async, since the message just need to be sendout and no need to wait for the response?

dilin-MS: Verified and remote the `setTimeout` machenism.

PR #853, opened 2019-11-29
Code refactor

dilin-MS:
1. Refactor telemetry worker to be singleton.
2. Remove redundant code.
3. Refactor command initialization in `extension.ts`.

### DIFF
--- a/src/Models/AzureUtility.ts
+++ b/src/Models/AzureUtility.ts
@@ -14,7 +14,7 @@ import {ConfigHandler} from '../configHandler';
 
 import {getExtension} from './Apis';
 import {ExtensionName} from './Interfaces/Api';
-import {TelemetryWorker, TelemetryContext, TelemetryResult} from '../telemetry';
+import {TelemetryWorker} from '../telemetry';
 import {EventNames} from '../constants';
 
 export interface ARMParameters {
--- a/src/Models/AzureUtility.ts
+++ b/src/Models/AzureUtility.ts
@@ -14,7 +14,7 @@ import {ConfigHandler} from '../configHandler';
 
 import {getExtension} from './Apis';
 import {ExtensionName} from './Interfaces/Api';
-import {TelemetryWorker, TelemetryContext} from '../telemetry';
+import {TelemetryWorker, TelemetryContext, TelemetryResult} from '../telemetry';
 import {EventNames} from '../constants';
 
 export interface ARMParameters {
@@ -431,17 +431,16 @@ export class AzureUtility {
       return undefined;
     }
 
-    const telemetryContext: TelemetryContext = {
-      properties: {
-        result: 'Succeeded',
-        error: '',
-        errorMessage: '',
-        subscription: subscription.description
-      },
-      measurements: {duration: 0}
-    };
+    const telemetryWorker = new TelemetryWorker(AzureUtility._context);
+    const telemetryContext = telemetryWorker.createContext();
+    telemetryContext.properties.subscription = subscription.description;
 
-    TelemetryWorker.sendEvent(EventNames.selectSubscription, telemetryContext);
+    try {
+      telemetryWorker.sendEvent(
+          EventNames.selectSubscription, telemetryContext);
+    } catch {
+      // If sending telemetry failed, skip the error to avoid blocking user.
+    }
     return subscription.description;
   }
 
--- a/src/Models/AzureUtility.ts
+++ b/src/Models/AzureUtility.ts
@@ -431,7 +431,7 @@ export class AzureUtility {
       return undefined;
     }
 
-    const telemetryWorker = new TelemetryWorker(AzureUtility._context);
+    const telemetryWorker = TelemetryWorker.getInstance(AzureUtility._context);
     const telemetryContext = telemetryWorker.createContext();
     telemetryContext.properties.subscription = subscription.description;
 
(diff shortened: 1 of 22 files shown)


src/telemetry.ts
```diff
@@ -1,137 +1,193 @@
...
-      await this._reporter.dispose();
-    }
-  }
+  private _reporter: TelemetryReporter|undefined;
+  private extensionContext: vscode.ExtensionContext|undefined;
 
-  static initialize(context: vscode.ExtensionContext): void {
-    const packageInfo = getPackageInfo(context);
-    if (!packageInfo) {
+  constructor(context: vscode.ExtensionContext) {
```
xinyiz1226: Why do we want to remove singleton from TelemetryWorker? I may be lack of context, let's discuss next week.

dilin-MS: Fixed to be singleton.

src/telemetry.ts
```diff
@@ -1,137 +1,200 @@
...
+   * Create telemetry context
+   */
+  createContext(): TelemetryContext {
+    const context: TelemetryContext = {properties: {}, measurements: {}};
+    context.properties.result = TelemetryResult.Succeeded;
+    context.properties.isInternal = TelemetryWorker.isInternalUser.toString();
+    if (this.extensionContext) {
+      context.properties.developEnvironment =
+          RemoteExtension.isRemote(this.extensionContext) ?
+          DevelopEnvironment.Container :
```
dooriya: VS Code Remote Development includes: remote SSH, WSL and container.
So we cannot infer from the remote environment that it is in a container environment.

src/telemetry.ts
```diff
@@ -1,137 +1,200 @@
...
+  measurements: {[key: string]: number};
 }
 
-export interface TelemetryMeasurements {
-  duration: number;
-  [key: string]: number;
-}
+export class TelemetryWorker {
+  private _reporter: TelemetryReporter|undefined;
+  private extensionContext: vscode.ExtensionContext|undefined;
```
dooriya: Please use consistent naming convention for the private field.

src/telemetry.ts
dilin-MS: Nice catch. Fixed.

src/telemetry.ts
dilin-MS: Fixed.

src/telemetry.ts
```diff
@@ -78,60 +55,146 @@ export class TelemetryWorker {
...
+    return context;
+  }
+
+  /**
+   * Send event telemetry
+   * @param eventName event name
+   * @param telemetryContext telemetry context
+   */
+  sendEvent(eventName: string, telemetryContext: TelemetryContext): void {
+    console.log(`[debug] Sending Telemetry. Event name： ${
```
dooriya: >[debug] Sending Telemetry. [](start = 17, length = 26)

Is that good to output debug info for prod release?

dilin-MS: Will remove it.

src/Models/AzureUtility.ts
```diff
@@ -431,17 +431,16 @@ export class AzureUtility {
...
-        subscription: subscription.description
-      },
-      measurements: {duration: 0}
-    };
+    const telemetryWorker = TelemetryWorker.getInstance(AzureUtility._context);
+    const telemetryContext = telemetryWorker.createContext();
+    telemetryContext.properties.subscription = subscription.description;
 
-    TelemetryWorker.sendEvent(EventNames.selectSubscription, telemetryContext);
+    try {
```
eriolchan: If we could put try...catch in TelemetryWorker.sendEvent so that we don't need to explicitly try...catch when calliing sendEvent?

src/telemetry.ts
```diff
@@ -78,60 +55,144 @@ export class TelemetryWorker {
...
+   */
+  sendEvent(eventName: string, telemetryContext: TelemetryContext): void {
+    if (!this._reporter) {
+      return;
+    }
+    if (telemetryContext) {
+      this._reporter.sendTelemetryEvent(
+          eventName, telemetryContext.properties,
+          telemetryContext.measurements);
+    } else {
```
eriolchan: According to @blackchoey, it is better to include "isInternal" property for each telemetry. It is helpful for BI to distinguish outside usage.

test/Device.test.ts
```diff
@@ -33,8 +28,8 @@ suite('IoT Device Workbench: Device', () => {
       const contextMock = new TestExtensionContext();
 
       const telemetryContext: TelemetryContext = {
-        properties: {result: 'Succeeded', error: '', errorMessage: ''},
-        measurements: {duration: 0}
+        properties: {result: TelemetryResult.Succeeded, error: '', errorMessage:
```
eriolchan: why not use telemetryWorker.createContext() here instead of manually construct

src/Models/AzureUtility.ts
dilin-MS: I will fix this part of job in another standalone PR focus on error handling.

src/telemetry.ts
dilin-MS: "isInternal" property is added when create a new telemetryContext. You can find it in the `createContext()` function.

test/Device.test.ts
dilin-MS: This part of code is commented out. Can we just delete it?

PR #856, opened 2019-12-02
Update test aikey

dilin-MS:
The origin aikey resource is removed from cloud. Create a new application insights service and replace the aikey.

### DIFF
--- a/scripts/modifyConfigIfRCVersion.js
+++ b/scripts/modifyConfigIfRCVersion.js
@@ -12,11 +12,13 @@ if (process.env.TRAVIS_TAG) {
     packageJson.name = testName;
     packageJson.displayName = testDisplayName;
     packageJson.publisher = testPublisher;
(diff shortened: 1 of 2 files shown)


scripts/modifyConfigIfRCVersion.js
```diff
@@ -12,11 +12,13 @@ if (process.env.TRAVIS_TAG) {
     packageJson.name = testName;
     packageJson.displayName = testDisplayName;
     packageJson.publisher = testPublisher;
+    packageJson.aiKey = process.env['TEST_AIKEY'];
```
xinyiz1226: do we want to check if AI Key is not a placeholder here?

dilin-MS: No. We will replace the "aikey" field in `package.json` with `TEST_AIKEY` if RC tag. And replace with `PROD_AIKEY` if production tag.

PR #859, opened 2019-12-02
fix issue of displaying incorrect title of settings

eriolchan:


src/constants.ts
```diff
@@ -2,19 +2,21 @@
 // Licensed under the MIT License.
 
 export class ConfigKey {
-  static readonly devicePath = 'DevicePath';
-  static readonly iotHubConnectionString = 'iothubConnectionString';
-  static readonly iotHubDeviceConnectionString = 'iothubDeviceConnectionString';
+  static readonly extensionName = 'vscode-iot-workbench';
```
dooriya: >extensionName = 'vscode-iot-workbench'; [](start = 18, length = 39)

Actually, the extension name for RC version is not 'vscode-iot-workbench'.

src/constants.ts
```diff
@@ -2,19 +2,21 @@
...
   static readonly eventHubConnectionPath = 'eventHubConnectionPath';
   static readonly functionAppId = 'functionAppId';
-  static readonly functionPath = 'FunctionPath';
-  static readonly boardId = 'BoardId';
+  static readonly functionPath = 'functionPath';
+  static readonly boardId = 'boardId';
   static readonly asaPath = 'StreamAnalyticsPath';
-  static readonly codeGeneratorVersion = 'IoTPnPCodeGenVersion';
-  static readonly projectType = 'ProjectType';
+  static readonly codeGeneratorVersion = 'iotPnpCodeGenVersion';
```
dooriya: Since you have change the some key names here: e.g. Board -> board. will that cause compatability issue for the old settings.json?

eriolchan:
For backward compatibility, not merge the change in this sprint. Send issue to VSCode to check if there is better way.

PR #861, opened 2019-12-03
[Code Refactor] Error handling

dilin-MS:
https://dev.azure.com/mseng/VSIoT/_workitems/edit/1652616

https://dev.azure.com/mseng/VSIoT/_workitems/edit/1648990

### DIFF
--- a/src/Models/ArduinoDeviceBase.ts
+++ b/src/Models/ArduinoDeviceBase.ts
@@ -141,10 +141,10 @@ export abstract class ArduinoDeviceBase implements Device {
     return true;
   }
 
-  abstract async create(): Promise<boolean>;
+  abstract async create(): Promise<void>;
 
   async createCore(board: Board|undefined, templateFiles: TemplateFileInfo[]):
-      Promise<boolean> {
+      Promise<void> {
     // Generate template files
     const createTimeScaffoldType = ScaffoldType.Local;
     if (!await FileUtility.directoryExists(
@@ -165,8 +165,6 @@ export abstract class ArduinoDeviceBase implements Device {
     // Configurate device environment
     await this.configDeviceEnvironment(
         this.deviceFolder, createTimeScaffoldType);
-
-    return true;
   }
 
   // Backward compatibility: Check configuration
--- a/src/Models/ArduinoDeviceBase.ts
+++ b/src/Models/ArduinoDeviceBase.ts
@@ -5,6 +5,7 @@ import * as fs from 'fs-plus';
 import * as path from 'path';
 import * as vscode from 'vscode';
 
+import {CancelOperationError} from '../CancelOperationError';
 import {Commands} from '../common/Commands';
 import {ConfigHandler} from '../configHandler';
 import {ConfigKey, DependentExtensions, FileNames, OperationType, OSPlatform, ScaffoldType} from '../constants';
@@ -162,10 +163,10 @@ export abstract class ArduinoDeviceBase implements Device {
     await this.generateCppPropertiesFile(createTimeScaffoldType, board);
 
     // Configurate device environment
-    const res = await this.configDeviceEnvironment(
+    await this.configDeviceEnvironment(
         this.deviceFolder, createTimeScaffoldType);
 
-    return res;
+    return true;
   }
 
   // Backward compatibility: Check configuration
@@ -317,7 +318,7 @@ export abstract class ArduinoDeviceBase implements Device {
   }
 
   async configDeviceEnvironment(
-      deviceRootPath: string, scaffoldType: ScaffoldType): Promise<boolean> {
+      deviceRootPath: string, scaffoldType: ScaffoldType): Promise<void> {
     if (!deviceRootPath) {
       throw new Error(
           'Unable to find the project device path, please open the folder and initialize project again.');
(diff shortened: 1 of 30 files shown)


src/Models/ArduinoDeviceBase.ts
```diff
@@ -162,10 +163,10 @@ export abstract class ArduinoDeviceBase implements Device {
     await this.generateCppPropertiesFile(createTimeScaffoldType, board);
 
     // Configurate device environment
-    const res = await this.configDeviceEnvironment(
+    await this.configDeviceEnvironment(
         this.deviceFolder, createTimeScaffoldType);
 
-    return res;
+    return true;
```
eriolchan: It seems that the function always returns true if there is no exception, then the return value is not necessary. Can we make the function return void instead of boolean?

src/Models/ContainerDeviceBase.ts
```diff
@@ -86,10 +86,10 @@ export abstract class ContainerDeviceBase implements Device {
     await this.generateTemplateFiles(
         createTimeScaffoldType, this.projectFolder, this.templateFilesInfo);
 
-    const res = await this.configDeviceEnvironment(
+    await this.configDeviceEnvironment(
         this.projectFolder, createTimeScaffoldType);
 
-    return res;
+    return true;
```
eriolchan: same question here.

src/Models/ContainerDeviceBase.ts
```diff
@@ -123,12 +123,11 @@ export abstract class ContainerDeviceBase implements Device {
     // Check remote
     const isRemote = RemoteExtension.isRemote(this.extensionContext);
     if (!isRemote) {
-      const res = await utils.askAndOpenInRemote(
+      await utils.askAndOpenInRemote(
           OperationType.Compile, this.channel, this.telemetryContext);
-      if (!res) {
-        return false;
-      }
+      return false;
```
eriolchan: why always return false here? Does caller consume the return boolean value?

src/Models/RaspberryPiDevice.ts
```diff
@@ -71,11 +72,9 @@ export class RaspberryPiDevice extends ContainerDeviceBase {
   async upload(): Promise<boolean> {
     const isRemote = RemoteExtension.isRemote(this.extensionContext);
     if (!isRemote) {
-      const res = await askAndOpenInRemote(
+      await askAndOpenInRemote(
           OperationType.Upload, this.channel, this.telemetryContext);
-      if (!res) {
-        return false;
-      }
+      return false;
```
eriolchan: same question here

src/constants.ts
```diff
@@ -66,8 +66,8 @@ export class FileNames {
 }
 
 export enum OperationType {
-  Compile = 'Compile device code',
-  Upload = 'Upload device code'
+  Compile = 'Device code compilation',
```
eriolchan: Is it better to express as verb+noun form? e.g. Compile device  code?

src/telemetry.ts
```diff
@@ -141,18 +141,19 @@ export class TelemetryWorker {
...
+      let isPopupErrorMsg = true;
       if (error instanceof CancelOperationError) {
         telemetryContext.properties.result = TelemetryResult.Cancelled;
+        isPopupErrorMsg = false;
       } else {
         telemetryContext.properties.result = TelemetryResult.Failed;
       }
-      ExceptionHelper.logError(outputChannel, error, true);
+      telemetryContext.properties.errorMessage = error.message;
+      ExceptionHelper.logError(outputChannel, error, isPopupErrorMsg);
```
eriolchan: It looks like CancelOperationError will show in output channel, do we still need to output message for CancelOperationError again in other function? e.g. CodeGeneratorCore

src/utils.ts
```diff
@@ -321,21 +319,15 @@ export async function askAndOpenInRemote(
...
+
     await vscode.commands.executeCommand('openindocker.reopenInContainer');
-  } else {
-    const message = `${operation} can only be executed in remote container.`;
-    channelShowAndAppendLine(channel, message);
-    telemetryContext.properties.errorMessage = `${operation} operation failed.`;
   }
 
-  return false;
+  throw new CancelOperationError(`${
```
eriolchan: Since the function return void, does it mean that it will always throw error here?

src/constants.ts
dilin-MS: The value is used for output. Now we have information message like `Device code compilation operation failed. Please do something.` That's why I change the string value to be a noun. It makes grammatical mistakes to say `Compile device code operation failed.`

src/Models/ArduinoDeviceBase.ts

src/Models/ContainerDeviceBase.ts
dilin-MS: Fixed.

src/Models/ContainerDeviceBase.ts
dilin-MS: The code below return true.
Here when not in remote, the compilation will fail and thus return false.

src/Models/RaspberryPiDevice.ts
dilin-MS: Same as above.

dilin-MS: If not remote, the operation failed. But will prompt to suggest user to open in remote.

src/telemetry.ts
dilin-MS: All cancel operation error print log in channel through this function, including CodeGeneratorCore.

src/utils.ts
dilin-MS: No need to throw cancel operation error here. Fixed.

PR #862, opened 2019-12-04
[Code refactor] workbench project create / load

dilin-MS:
Code refactor:
1. Break long functions into small functions.
2. Set `projectRootPath`, `deviceRootPath`, `iotworkbenchProjectFile` as class properties. Change the way to construct and create project.
3. Add comments.
4. Rename functions for better readbility.
5. Remove useless code.

### DIFF
--- a/src/FileUtility.ts
+++ b/src/FileUtility.ts
@@ -109,15 +109,15 @@ export class FileUtility {
   /**
    * Convert Json object to Json string and write into target file path.
    * @param type Scaffold type
-   * @param filePath tartet file path
+   * @param fileDestPath tartet file path
    * @param data Json object
    */
   static async writeJsonFile(
       // tslint:disable-next-line: no-any
-      type: ScaffoldType, filePath: string, data: any): Promise<void> {
+      type: ScaffoldType, fileDestPath: string, data: any): Promise<void> {
     const indentationSpace = 4;
     const jsonString = JSON.stringify(data, null, indentationSpace);
-    await FileUtility.writeFile(type, filePath, jsonString);
+    await FileUtility.writeFile(type, fileDestPath, jsonString);
   }
 
   static async readFile(
--- a/src/Models/IoTContainerizedProject.ts
+++ b/src/Models/IoTContainerizedProject.ts
@@ -11,7 +11,6 @@ import {FileUtility} from '../FileUtility';
 import {TelemetryContext, TelemetryResult, TelemetryWorker} from '../telemetry';
 import {channelShowAndAppendLine, getFirstWorkspaceFolderPath} from '../utils';
 
-import {Component} from './Interfaces/Component';
 import {ProjectHostType} from './Interfaces/ProjectHostType';
 import {ProjectTemplateType, TemplateFileInfo} from './Interfaces/ProjectTemplate';
 import {IoTWorkbenchProjectBase, OpenScenario} from './IoTWorkbenchProjectBase';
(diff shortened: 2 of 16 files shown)


src/FileUtility.ts
```diff
@@ -109,15 +109,15 @@ export class FileUtility {
   /**
    * Convert Json object to Json string and write into target file path.
    * @param type Scaffold type
-   * @param filePath tartet file path
+   * @param fileDestPath tartet file path
```
eriolchan: It is ok to just use filePath as parameter name since function writeJsonFile indicate the filePath is target file path to write content.

src/Models/IoTContainerizedProject.ts
```diff
@@ -23,148 +24,109 @@ const raspberryPiDeviceModule =
...
+      telemetryContext: TelemetryContext, rootFolderPath?: string) {
     super(context, channel, telemetryContext);
     this.projectHostType = ProjectHostType.Container;
-  }
 
-  async load(scaffoldType: ScaffoldType, initLoad = false): Promise<boolean> {
-    if (!(vscode.workspace.workspaceFolders &&
-          vscode.workspace.workspaceFolders.length > 0)) {
-      return false;
+    if (rootFolderPath) {
```
eriolchan: would it be better as 
```
this.projectRootPath = rootFolderPath || getFirstWorkspaceFolderPath();
if (!this.projectRootPath) {
  throw new Error(`Fail to get root folder.`);
}
```

src/Models/IoTContainerizedProject.ts
```diff
@@ -23,142 +24,104 @@ const raspberryPiDeviceModule =
...
+          `Workspace config file path is empty. Please initialize the project first.`);
+    }
+    await FileUtility.writeJsonFile(
+        createTimeScaffoldType, this.iotWorkbenchProjectFilePath,
+        projectConfig);
+
+    // Check components prerequisites
+    this.componentList.forEach(async item => {
+      const res = await item.checkPrerequisites();
+      if (!res) {
```
eriolchan: do we need to throw error if prerequisites is not met?

src/Models/IoTContainerizedProject.ts
```diff
@@ -167,13 +130,14 @@ export class IoTContainerizedProject extends IoTWorkbenchProjectBase {
...
   async openProject(
-      projectPath: string, openInNewWindow: boolean,
-      openScenario: OpenScenario) {
-    if (!FileUtility.directoryExists(ScaffoldType.Local, projectPath)) {
-      channelShowAndAppendLine(
-          this.channel, `Can not find project path ${projectPath}.`);
-      return;
+      scaffoldType: ScaffoldType, openInNewWindow: boolean,
+      openScenario: OpenScenario): Promise<void> {
+    if (!await FileUtility.directoryExists(
```
eriolchan: The root path validation logic is spread in many function, can we abstract it as a common function to validateRootPath?

src/Models/IoTWorkbenchProjectBase.ts
```diff
@@ -338,12 +331,30 @@ export abstract class IoTWorkbenchProjectBase {
...
+          `Update ${FileNames.iotworkbenchprojectFileName} file failed: ${
+              error.message}`);
+    }
+  }
+
+  /**
+   * Get project configs from iot workbench project file
+   * @param type Scaffold type
+   */
+  async getProjectConfig(type: ScaffoldType): Promise<{[key: string]: string}> {
```
eriolchan: Since the function return JSON object, can we just return any instead of{[key:string]:string} ?

src/utils.ts
```diff
@@ -457,16 +461,12 @@ export async function handleExternalProject(
       return;
```
eriolchan: Since the return value is string, we can just return empty string ("") instead of undefined.

src/utils.ts
```diff
@@ -569,10 +569,8 @@ export async function constructAndLoadIoTProject(
...
     }
 
-    const result = await iotProject.load(scaffoldType);
-    if (!result) {
-      throw new Error(`Failed to load project.`);
-    }
+    await iotProject.load(scaffoldType);
```
eriolchan: why there are both project and iotProject?

src/Models/IoTContainerizedProject.ts
```diff
@@ -23,142 +24,104 @@ const raspberryPiDeviceModule =
...
+      const firstWorkspaceFolder = getFirstWorkspaceFolderPath();
+      if (!firstWorkspaceFolder) {
+        throw new Error(`Fail to get first workspace folder.`);
+      }
+      this.projectRootPath = firstWorkspaceFolder;
     }
 
-    this.projectRootPath = vscode.workspace.workspaceFolders[0].uri.fsPath;
-
-    await this.generateOrUpdateIotWorkbenchProjectFile(
```
eriolchan: after refactor, where to trigger create/update and init device now?

src/FileUtility.ts
dilin-MS: Yes. Just a enhancement to make it more apparent.

src/Models/IoTContainerizedProject.ts
dilin-MS: LGTM. Fixed.

src/Models/IoTContainerizedProject.ts
dilin-MS: Yes. Fixed.

src/Models/IoTContainerizedProject.ts

src/utils.ts
dilin-MS: The return value is Promise<void>, not string.

src/Models/IoTWorkbenchProjectBase.ts
dilin-MS: Fixed.

PR #869, opened 2019-12-10
Configure project environment command flow adjustment

dilin-MS:
https://dev.azure.com/mseng/VSIoT/_workitems/edit/1652089

### DIFF
--- a/src/ProjectEnvironmentConfiger.ts
+++ b/src/ProjectEnvironmentConfiger.ts
@@ -85,7 +85,7 @@ export class ProjectEnvironmentConfiger {
           context, channel, telemetryContext);
       if (!project) {
         // Ensure the project is correctly open.
-        await utils.handleIncorrectlyOpenedIoTWorkspaceProject(
+        await utils.ensureIoTWorkspaceProjectIsCorrectlyOpened(
             telemetryContext);
       }
       telemetryContext.properties.projectHostType = ProjectHostType.Workspace;
--- a/src/ProjectEnvironmentConfiger.ts
+++ b/src/ProjectEnvironmentConfiger.ts
@@ -2,8 +2,7 @@
 // Licensed under the MIT License.
 
 'use strict';
-// The module 'vscode' contains the VS Code extensibility API
-// Import the module and reference it with the alias vscode in your code below
+
 import * as vscode from 'vscode';
 import * as utils from './utils';
 
@@ -12,7 +11,7 @@ import {ScaffoldType, PlatformType} from './constants';
 import {RemoteExtension} from './Models/RemoteExtension';
 import {IoTWorkbenchProjectBase, OpenScenario} from './Models/IoTWorkbenchProjectBase';
 import {ProjectHostType} from './Models/Interfaces/ProjectHostType';
-import {CancelOperationError} from './CancelOperationError';
+import {configExternalCMakeProjectToIoTContainerProject} from './utils';
 
 const impor = require('impor')(__dirname);
 const ioTWorkspaceProjectModule = impor('./Models/IoTWorkspaceProject') as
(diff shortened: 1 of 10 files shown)


src/ProjectEnvironmentConfiger.ts
```diff
@@ -78,54 +66,48 @@ export class ProjectEnvironmentConfiger {
...
       }
-    }
 
-    let project;
-    if (platform === PlatformType.EmbeddedLinux) {
+      project = new ioTWorkspaceProjectModule.IoTWorkspaceProject(
+          context, channel, telemetryContext);
+      if (!project) {
+        // Ensure the project is correctly open.
+        await utils.handleIncorrectlyOpenedIoTWorkspaceProject(
```
eriolchan: How about name the function as "ensureProjectIsOpen" to make method name easy to understand?

src/utils.ts
```diff
@@ -478,6 +479,95 @@ export async function handleExternalProject(
...
+/**
+ * Config External CMake Project config file as an IoT Workbench Container
+ * Project. Throw cancel operation error if not CMake project. Update project
+ * host type and board id in IoT Workbench project file.
+ * @param scaffoldType
+ */
+export async function configExternalCMakeProjectToIoTContainerProject(
+    scaffoldType: ScaffoldType): Promise<void> {
+  const projectRootPath = getFirstWorkspaceFolderPath();
+  // Check if cmake project
```
eriolchan: the comment message is not complete?

src/utils.ts
```diff
@@ -478,6 +479,95 @@ export async function handleExternalProject(
...
+  const cmakeFile = path.join(projectRootPath, FileNames.cmakeFileName);
+  if (!await FileUtility.fileExists(scaffoldType, cmakeFile)) {
+    const message = `Missing ${
+        FileNames.cmakeFileName} to be configured as Embedded Linux project.`;
+    vscode.window.showWarningMessage(message);
+    throw new CancelOperationError(message);
+  }
+
+  const iotworkbenchprojectFile =
```
eriolchan: can we make "iotworkbenchproject" as camel case to better read?

src/utils.ts
```diff
@@ -478,6 +479,95 @@ export async function handleExternalProject(
...
+    const projectConfig =
+        await getProjectConfig(type, iotWorkbenchProjectFilePath);
+
+    // Update project host type
+    projectConfig[`${ConfigKey.projectHostType}`] =
+        ProjectHostType[projectHostType];
+
+    // Add config version for easier backward compatibility in the future.
+    const workbenchVersion = '1.0.0';
+    projectConfig[`${ConfigKey.workbenchVersion}`] = workbenchVersion;
```
eriolchan: Do we always override workbench version or update it when it is undefined?

src/utils.ts
dilin-MS: It should be "Check if it is a cmake project". I omitted the "it is a".
Will fix it.

src/utils.ts
dilin-MS: The statement is meant to update version if null. Currently only one version. 
Will fix this.

src/ProjectEnvironmentConfiger.ts
dilin-MS: Fixed.

src/utils.ts

src/utils.ts
dilin-MS: The former "iotworkbenchProjectFile" uses this case because it wants to represent the "**`.iotworkbenchproject` file**", in which case it is camel case. Since we have two kinds of reperesentation "*iotworkbenchprojectFile*" and "*iotWorkbenchProjectFile*", I will delete the former one and use "*iotWorkbenchProjectFile*" for consolidation.

src/utils.ts

src/utils.ts

PR #873, opened 2019-12-12
Fix situation when iot workspace file has different config file name

dilin-MS:


src/Models/IoTWorkspaceProject.ts
```diff
@@ -185,14 +185,38 @@ export class IoTWorkspaceProject extends IoTWorkbenchProjectBase {
...
 
-  async openProject(
-      scaffoldType: ScaffoldType, openInNewWindow: boolean,
-      openScenario: OpenScenario): Promise<void> {
+  /**
+   * Validate workspace config file exists.
+   * Handles situation when workspace config file has a different name.
+   * @param scaffoldType
+   */
+  async validateWorkspaceConfigFilePath(scaffoldType: ScaffoldType):
```
eriolchan: It seems this function is try to assign workspace file to variable this.workspaceConfigFilePath? The logic is to find files with specified suffix in folder. It throws error if there is no file found, what if there are multiple file found?
BTW, you can use this method (vscode.workspace.findFiles) to find files.

dilin-MS: If there is many, we use the first file found.

src/Models/IoTWorkspaceProject.ts
```diff
@@ -185,14 +185,38 @@ export class IoTWorkspaceProject extends IoTWorkbenchProjectBase {
...
+      Promise<void> {
     if (!await FileUtility.fileExists(
             scaffoldType, this.workspaceConfigFilePath)) {
-      throw new Error(`Workspace config file ${
-          this.workspaceConfigFilePath} does not exist. Please initialize the project first.`);
+      // Handles situation when workspace config file has a different name.
```
dooriya: Do not assume that the workspace file name is the same as the root folder name.

src/Models/IoTWorkspaceProject.ts
```diff
@@ -185,14 +185,38 @@ export class IoTWorkspaceProject extends IoTWorkbenchProjectBase {
...
-      openScenario: OpenScenario): Promise<void> {
+  /**
+   * Validate workspace config file exists.
+   * Handles situation when workspace config file has a different name.
+   * @param scaffoldType
+   */
+  async validateWorkspaceConfigFilePath(scaffoldType: ScaffoldType):
+      Promise<void> {
     if (!await FileUtility.fileExists(
             scaffoldType, this.workspaceConfigFilePath)) {
```
dooriya: We may need to revisit the APIs in FileUtility, I don't think a general file utility API should care about the scaffold type.

PR #880, opened 2019-12-19
Fix auto popup landing page

dilin-MS:
1. Refactor IoTWorkbenchSettings to be singleton.
2. Replace os string 'win32'/'linux'/'darwin' with constants.
3. Fix auto pop up landing page: Only pop up when user first install iot workbench extension.

### DIFF
--- a/src/IoTSettings.ts
+++ b/src/IoTSettings.ts
@@ -3,47 +3,36 @@
 
 import * as path from 'path';
 import * as vscode from 'vscode';
-import * as sdk from 'vscode-iot-device-cube-sdk';
 
 import {CancelOperationError} from './CancelOperationError';
 import {ConfigHandler} from './configHandler';
-import {OSPlatform} from './constants';
+import {ConfigKey, OSPlatform} from './constants';
 import {PickWithData} from './Models/Interfaces/UI';
+import {getHomeDir, getPlatform} from './utils';
 
 export class IoTWorkbenchSettings {
-  private _workbenchPath = '';
+  private workbenchPath = '';
+  private static instance: IoTWorkbenchSettings|undefined;
 
   private constructor() {}
 
-  static async createAsync() {
-    const iotWorkbenchSettings = new IoTWorkbenchSettings();
+  static async getInstance() {
+    if (!this.instance) {
+      this.instance = new IoTWorkbenchSettings();
+      this.instance.workbenchPath =
+          ConfigHandler.get<string>(ConfigKey.workbench) ||
+          (await this.getDefaultWorkbenchPath());
+      await ConfigHandler.update(
+          ConfigKey.workbench, this.instance.workbenchPath,
+          vscode.ConfigurationTarget.Global);
+    }
 
-    iotWorkbenchSettings._workbenchPath =
-        await IoTWorkbenchSettings.getWorkbenchPath();
-
-    return iotWorkbenchSettings;
+    return this.instance;
   }
 
-  static async getPlatform(): Promise<string> {
-    const localOs = sdk.Utility.require('os') as typeof import('os');
-    const getPlatform = await localOs.platform;
+  static async getDefaultWorkbenchPath(): Promise<string> {
     const platform = await getPlatform();
-    return platform;
-  }
-
-  static async getOs(): Promise<string> {
-    const localOs = sdk.Utility.require('os') as typeof import('os');
-    const getHomeDir = await localOs.homedir;
     const homeDir = await getHomeDir();
-    return homeDir;
-  }
-
-  static async getWorkbenchPath(): Promise<string> {
-    const localOs = sdk.Utility.require('os') as typeof import('os');
-    const getHomeDir = await localOs.homedir;
-    const homeDir = await getHomeDir();
-    const getPlatform = await localOs.platform;
-    const platform = await getPlatform();
 
     let _workbenchPath = '';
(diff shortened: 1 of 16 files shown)


src/IoTSettings.ts
```diff
@@ -3,47 +3,36 @@
...
 
   private constructor() {}
 
-  static async createAsync() {
-    const iotWorkbenchSettings = new IoTWorkbenchSettings();
-
-    iotWorkbenchSettings._workbenchPath =
-        await IoTWorkbenchSettings.getWorkbenchPath();
+  static async getInstance() {
+    if (!this.instance) {
```
eriolchan: Since instance is a static variable, it is better to use IoTWorkbenchSettings.instance instead of this.instance.

src/IoTSettings.ts
```diff
@@ -3,47 +3,36 @@
...
+      this.instance = new IoTWorkbenchSettings();
+      this.instance.workbenchPath =
+          ConfigHandler.get<string>(ConfigKey.workbench) ||
+          (await this.getDefaultWorkbenchPath());
+      await ConfigHandler.update(
```
eriolchan: Do we really need to update config again if Configkey.workbench is already set? I guess you mean update the config when it is not set?

src/IoTSettings.ts
```diff
@@ -3,47 +3,36 @@
...
+  private static instance: IoTWorkbenchSettings|undefined;
 
   private constructor() {}
 
-  static async createAsync() {
-    const iotWorkbenchSettings = new IoTWorkbenchSettings();
-
-    iotWorkbenchSettings._workbenchPath =
-        await IoTWorkbenchSettings.getWorkbenchPath();
+  static async getInstance() {
```
eriolchan: You could add return type in method signature as other method.

src/IoTSettings.ts
```diff
@@ -59,21 +48,12 @@ export class IoTWorkbenchSettings {
...
-      return this._workbenchPath;
-    }
+  getWorkbenchPath(): string {
+    return ConfigHandler.get<string>(ConfigKey.workbench) || this.workbenchPath;
   }
 
-  async setWorkbenchPath(showMessage = true) {
-    let userWorkbenchPath: string|undefined =
-        ConfigHandler.get<string>('workbench') || this._workbenchPath;
+  async setWorkbenchPath(showMessage = true): Promise<void> {
```
eriolchan: In which case, we would call this function with showMessage=false? I am not sure if the parameter is necessary.

eriolchan: The function looks like asking user to select workbench path, would selectWorkbenchPath a better name?

src/IoTSettings.ts
```diff
@@ -101,22 +81,22 @@ export class IoTWorkbenchSettings {
...
           throw new CancelOperationError('Change workbench cancelled.');
         }
-        return userWorkbenchPath;
+        return;
       }
     } else if (selection !== undefined) {
       userWorkbenchPath = selection.data;
     } else {
-      userWorkbenchPath = undefined;
+      userWorkbenchPath = '';
```
eriolchan: I suggest to re-organize this function logic to make it clear, such as
1. show quick pick to ask user to select workbench folder.
2. if(!selection), return. (do we need to throw cancel error?)
3. if selection.data === '$', show open dialog for further selection.
4. else set the userWorkbenchPath value.
5. update config.

src/Models/Esp32Device.ts
```diff
@@ -42,7 +41,7 @@ export class Esp32Device extends ArduinoDeviceBase {
     let packageRootPath = '';
     let version = '0.0.1';
 
-    if (plat === 'win32') {
+    if (plat === OSPlatform.WIN32) {
```
eriolchan: It is better to align the variable name from plat to platform.

src/IoTSettings.ts
dilin-MS: Fixed.

src/IoTSettings.ts
dilin-MS: Yes. Update user setting when it is not set.

src/IoTSettings.ts
dilin-MS: Will remove it since currently no scenario will set `showMessage` as false.

dilin-MS: Select workbench path is just one step of it. This function is meant to update workbench path.

src/Models/Esp32Device.ts
dilin-MS: Fixed.

src/IoTSettings.ts
dilin-MS: Fixed.

src/IoTSettings.ts
```diff
@@ -66,37 +85,25 @@ export class IoTWorkbenchSettings {
...
+  private async selectFolder(): Promise<string> {
+    const options: vscode.OpenDialogOptions = {
+      canSelectMany: false,
+      openLabel: 'Select',
+      canSelectFolders: true,
+      canSelectFiles: false
+    };
+
+    const folderUri = await vscode.window.showOpenDialog(options);
+    if (!(folderUri && folderUri.length > 0)) {
```
eriolchan: this would be better to understand
```
if (!folderUri || folderUri.length === 0)
```

dilin-MS: Fixed.

PR #887, opened 2019-12-26
Bugs fix: auto configure; worspace config file name

dilin-MS:
1. Fix auto-configure project env bug: when iot workbench project does not have ".vscode/tasks.json" file, pop up to ask to configure project environment and auto configure according to project platform type.
2. Fix workspace config file name bug: worspace config file path will be initialized in `load()` and `create()` methods respectively.
3. Fix typo

### DIFF
--- a/src/Models/IoTWorkspaceProject.ts
+++ b/src/Models/IoTWorkspaceProject.ts
@@ -10,7 +10,7 @@ import {ConfigHandler} from '../configHandler';
 import {ConfigKey, EventNames, FileNames, ScaffoldType} from '../constants';
 import {FileUtility} from '../FileUtility';
 import {TelemetryContext, TelemetryWorker} from '../telemetry';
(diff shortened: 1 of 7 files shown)


src/Models/IoTWorkspaceProject.ts
```diff
@@ -86,13 +73,26 @@ export class IoTWorkspaceProject extends IoTWorkbenchProjectBase {
...
 
+    // Init and update iot workbench project file
     this.iotWorkbenchProjectFilePath =
         path.join(this.deviceRootPath, FileNames.iotWorkbenchProjectFileName);
-
-    // Update iot workbench project file.
     await updateProjectHostTypeConfig(
         scaffoldType, this.iotWorkbenchProjectFilePath, this.projectHostType);
 
+    // Init workspace config file
```
eriolchan: I remember the same logic is shared by other place, you could extract as a method later

dilin-MS: Okay.

PR #893, opened 2019-12-30
Fix nightly build settings

dilin-MS:


.travis.yml
```diff
@@ -21,7 +21,9 @@ env:
     - ISPRODTAG=^v?[0-9]+\.[0-9]+\.[0-9]+$
     - ISTESTTAG=^v?[0-9]+\.[0-9]+\.[0-9]+-[rR][cC]
     - RC_EXTENSION_ID="iotdevexbuild.test-owl-project"
-    - NIGHTLY_BUILD_ID="vsciot-vscode.vscode-iot-workbench-nightly"
+    - NIGHTLY_BUILD_ID="iotdevexbuild.vscode-iot-workbench-nightly"
```
dooriya: it should be renamed to test-owl-project-nightly

dilin-MS: Fixed.

.travis.yml
```diff
@@ -72,7 +77,7 @@ deploy:
       tags: true
       all_branches: true
       # if it's a PROD tag (something like 1.0.0), then publish extension to market.
-      condition: "$TRAVIS_TAG =~ $ISPRODTAG"
+      condition: "$IS_PROD = true"
```
eriolchan: Is the value of $IS_PROD calculated or configured in env variable?

.travis.yml
```diff
@@ -57,6 +61,7 @@ deploy:
   # deploy nightly build to github release
   - provider: releases
     api_key: $GIT_TOKEN
+    prerelease: true
```
dooriya: >prerelease: true [](start = 4, length = 16)

Keep consistency to use the $IS_TEST variable if it is used elsewhere

.travis.yml
dilin-MS: Fxied. Nice catch.

.travis.yml
dilin-MS: For the "deploy nightly build to github release" deployment, `prerelease` tag will always be true.
For the "deploy to github release" deployment, it can be production release or RC version release (which will have `$IS_TEST == true`). `prerelease` tag will only be true when $IS_TEST is true.

PR #895, opened 2019-12-30
Error Handling Refactor

dilin-MS:


src/DigitalTwin/CodeGeneratorCore.ts
```diff
@@ -23,7 +23,7 @@ import {RemoteExtension} from '../Models/RemoteExtension';
 import {DigitalTwinUtility} from './DigitalTwinUtility';
 import {CodeGenUtility} from './DigitalTwinCodeGen/CodeGenUtility';
 import {FileUtility} from '../FileUtility';
-import {CancelOperationError} from '../CancelOperationError';
+import {CancelOperationError} from '../common/CancelOperationError';
```
xinyiz1226: It is good to move this to common folder, and I think it will be better to put all Error class definition in one file, so can we rename it for general purpose like Errors.ts?

src/Models/ArduinoDeviceBase.ts
```diff
@@ -232,41 +233,28 @@ export abstract class ArduinoDeviceBase implements Device {
...
-  async generateCrc(
-      context: vscode.ExtensionContext, channel: vscode.OutputChannel) {
+  async generateCrc(channel: vscode.OutputChannel): Promise<void> {
     if (!(vscode.workspace.workspaceFolders &&
           vscode.workspace.workspaceFolders.length > 0)) {
-      const message = 'No workspace opened.';
-      vscode.window.showWarningMessage(message);
-      utils.channelShowAndAppendLine(channel, message);
-      return false;
+      throw new Error('No workspace opened.');
```
xinyiz1226: can we define a new class for this kind of error? There are several errors related to environment like workspace not opened, or device path not found. also we may need to provide suggestions to users how they can fix it.

src/DigitalTwin/CodeGeneratorCore.ts
xinyiz1226: there is a typo in CancelOperationError class: "Openration cancelled" -> "Operation cancelled". Can you help fix this in this PR as well?

src/Models/ArduinoDeviceBase.ts
```diff
@@ -232,41 +233,28 @@ export abstract class ArduinoDeviceBase implements Device {
...
-      vscode.window.showWarningMessage(message);
-      utils.channelShowAndAppendLine(channel, message);
-      return false;
+      throw new Error('No device path found in workspace configuration.');
     }
     const deviceBuildLocation = path.join(
         vscode.workspace.workspaceFolders[0].uri.fsPath, '..', devicePath,
         '.build');
 
     if (!deviceBuildLocation) {
```
xinyiz1226: Will this variable be null?

src/Models/ArduinoDeviceBase.ts
```diff
@@ -232,41 +233,28 @@ export abstract class ArduinoDeviceBase implements Device {
...
     const devicePath = ConfigHandler.get<string>(ConfigKey.devicePath);
     if (!devicePath) {
-      const message = 'No device path found in workspace configuration.';
-      vscode.window.showWarningMessage(message);
-      utils.channelShowAndAppendLine(channel, message);
-      return false;
+      throw new Error('No device path found in workspace configuration.');
     }
     const deviceBuildLocation = path.join(
         vscode.workspace.workspaceFolders[0].uri.fsPath, '..', devicePath,
```
xinyiz1226: what will happen if workspaceFolders[0] does not exist?

src/Models/ArduinoDeviceBase.ts
```diff
@@ -232,41 +233,28 @@ export abstract class ArduinoDeviceBase implements Device {
...
     }
 
     const binFiles = fs.listSync(deviceBuildLocation, ['bin']);
     if (!binFiles || !binFiles.length) {
-      const message =
-          'No bin file found. Please run the command of Device Compile first.';
-      vscode.window.showWarningMessage(message);
-      utils.channelShowAndAppendLine(channel, message);
-      return false;
+      throw new Error(
```
xinyiz1226: define new class ResourceNotFoundError and use here

src/Models/ArduinoDeviceBase.ts
```diff
@@ -288,14 +276,14 @@ export abstract class ArduinoDeviceBase implements Device {
...
-        return false;
+        throw new CancelOperationError('Bin file selection cancelled.');
       }
 
       binFilePath = choice.description;
     }
 
     if (!binFilePath || !fs.existsSync(binFilePath)) {
-      return false;
+      throw new Error(`Bin file path does not exist: ${binFilePath}.`);
```
xinyiz1226: reuse ResourceNotFoundError here

src/Models/RemoteExtension.ts
```diff
@@ -49,16 +49,13 @@ export class RemoteExtension {
...
+   * If in remote environment, throw error.
    */
-  static checkLocalBeforeRunCommand(context: vscode.ExtensionContext): boolean {
+  static ensureLocalBeforeRunCommand(context: vscode.ExtensionContext): void {
     if (RemoteExtension.isRemote(context)) {
-      const message =
-          `The command is not supported to be run in a remote environment. Open a new window and run this command again.`;
-      vscode.window.showWarningMessage(message);
-      return false;
+      throw new Error(
```
xinyiz1226: Define RemoteEnvNotSupportedError and use here

src/DigitalTwin/CodeGeneratorCore.ts
dilin-MS: The typo is fixed in `pre-release-0.11.0-rc` branch (see [commit](https://github.com/microsoft/vscode-iot-workbench/commit/529944efd1374d493d0c66d5d01d63adcb192525#diff-98676cbf9c6148ad805552a7eed6c0c3)). Later this branch will be merged back to develop branch.

I will double check this typo after the prerelease branch merge back into develop branch.

dilin-MS: Currently we have only one special error type: the `CancelOperationError`. If we have more error types, I will make this file a more general one.

src/Models/RemoteExtension.ts
dilin-MS: Done.

src/Models/AZ3166Device.ts
```diff
@@ -160,93 +241,17 @@ export class AZ3166Device extends ArduinoDeviceBase {
...
-      const res = await this.setDeviceConfig(UDS, ConfigDeviceOptions.UDS);
-      if (!res) {
-        return false;
-      } else {
-        vscode.window.showInformationMessage(
-            'Configure Unique Device String (UDS) completed successfully.');
-        return true;
-      }
-    }
+    return configSelection.detail;
```
dooriya: > configSelection.detail [](start = 10, length = 23)

Suggest using 'label' of the selection instead of the 'details', which is usually a long description string of the selected option. 

Since you've define an Enum for the device setting options, you can convert the selection into the 'ConfigDeviceOptions' enum, and use enum instead of string in the subsequent steps

src/Models/AZ3166Device.ts
```diff
@@ -150,8 +157,82 @@ export class AZ3166Device extends ArduinoDeviceBase {
...
+    }
+
+    vscode.window.showInformationMessage(
+        `Successfully configure ${deviceSettingsType}.`);
+  }
+
+  /**
+   * Select device setting type.
+   */
+  private async selectDeviceSettingType(): Promise<string|undefined> {
```
dooriya: >Promise<string|undefined [](start = 43, length = 24)

Why not using the 'ConfigDeviceOptions' enum

src/Models/AZ3166Device.ts
```diff
@@ -150,8 +157,82 @@ export class AZ3166Device extends ArduinoDeviceBase {
...
+    let deviceSettingsType = '';  // For log info
+    switch (deviceSettingType) {
+      case ConfigDeviceOptions.CRC:
+        await this.generateCrc(this.channel);
+        return;
+      case ConfigDeviceOptions.ConnectionString:
+        // Get device connection string
+        credentials = await this.getDeviceConnectionString();
+
+        deviceSettingsType = 'device connection string';
```
dooriya: >deviceSettingsType = 'device connection string'; [](start = 8, length = 48)

Please use the enum string you've already defined for it.

src/Models/RaspberryPiDevice.ts
```diff
@@ -306,7 +264,7 @@ export class RaspberryPiDevice extends ContainerDeviceBase {
     };
     let raspiPassword = await vscode.window.showInputBox(raspiPasswordOption);
     if (raspiPassword === undefined) {
```
eriolchan: could use ! for falsy value check.
```
if (!raspiPassword)
```

src/utils.ts
```diff
@@ -127,11 +127,9 @@ export function getFirstWorkspaceFolderPath(): string {
   if (!(vscode.workspace.workspaceFolders &&
```
eriolchan: Same here to apply inverted into condition statement. Is uri.fsPath possible to be undefined?
```
if (!vscode.workspace.workspaceFolders || vscode.workspace.workspaceFolders.length === 0)
```

src/utils.ts
```diff
@@ -587,7 +585,13 @@ export async function properlyOpenIoTWorkspaceProject(
...
-  const rootPath = getFirstWorkspaceFolderPath();
+  let rootPath = '';
+  try {
+    rootPath = getFirstWorkspaceFolderPath();
+  } catch (error) {
+    return false;
+  }
+
   const workbenchFileName =
       path.join(rootPath, 'Device', FileNames.iotWorkbenchProjectFileName);
```
eriolchan: 'Device' could be a constant of FolderNames

src/utils.ts
```diff
@@ -608,7 +612,16 @@ export async function constructAndLoadIoTProject(
     telemetryContext: TelemetryContext, isTriggeredWhenExtensionLoad = false) {
   const scaffoldType = ScaffoldType.Workspace;
 
-  const projectFileRootPath = getFirstWorkspaceFolderPath();
+  let projectFileRootPath = '';
+  try {
+    projectFileRootPath = getFirstWorkspaceFolderPath();
+  } catch (error) {
+    if (isTriggeredWhenExtensionLoad) {
```
eriolchan: Why the behavior is different when isTriggeredWhenExtensionLoad is true? what is that scenario?

src/DigitalTwin/CodeGeneratorCore.ts
dilin-MS: Will do that after move to eslint.

src/utils.ts
dilin-MS: When a vscode window is open, before user trigger any workbench command, we will try load the current folder if any. In this situation, it is the `isTriggeredWhenExtensionLoad == true` scenario. In this situation, we do not want to throw error when try to load project since user may not want to use workbench at all.

src/utils.ts

src/Models/RaspberryPiDevice.ts
dilin-MS: Since password can be empty string. We use `raspiPassword === undefined` instead of `!raspiPassword` here.

src/Models/ArduinoDeviceBase.ts
```diff
@@ -232,41 +233,28 @@ export abstract class ArduinoDeviceBase implements Device {
     }
   }
 
-  async generateCrc(
-      context: vscode.ExtensionContext, channel: vscode.OutputChannel) {
+  async generateCrc(channel: vscode.OutputChannel): Promise<void> {
     if (!(vscode.workspace.workspaceFolders &&
```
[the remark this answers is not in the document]

src/utils.ts

src/Models/AZ3166Device.ts
dilin-MS: Fixed.

src/Models/AZ3166Device.ts
dilin-MS: Fixed.

src/Models/AZ3166Device.ts
dilin-MS: Fixed.

src/Models/ArduinoDeviceBase.ts
dilin-MS: Should check the directory exists instead of checking whether the path is empty. Fixed.

src/Models/ArduinoDeviceBase.ts

src/Models/ArduinoDeviceBase.ts

src/Models/ArduinoDeviceBase.ts
```diff
@@ -310,15 +293,14 @@ export abstract class ArduinoDeviceBase implements Device {
...
-
-    return true;
   }
 
   async configDeviceEnvironment(
       deviceRootPath: string, scaffoldType: ScaffoldType): Promise<void> {
     if (!deviceRootPath) {
-      throw new Error(
-          'Unable to find the project device path, please open the folder and initialize project again.');
+      throw new AugumentEmptyOrNullError(
```
xinyiz1226: should be "Argument"

src/common/Error/Error.ts
```diff
@@ -0,0 +1,121 @@
...
+export class InternalError extends Error {
+  constructor(message: string) {
+    super(`Internal Error: ${message}.`);
+    this.name = 'InternalError';
+  }
+}
+
+export class AugumentEmptyOrNullError extends Error {
+  constructor(augument: string, suggestedOperation?: string) {
+    super(`Augment ${augument} is empty or null. ${suggestedOperation}`);
```
xinyiz1226: change to "Argument"

src/common/Error/Error.ts
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Error class used when user cancel operation.
+ */
+export class OperationCanceledError extends Error {
+  constructor(message: string) {
+    super(`Operation cancelled: ${message}`);
+    this.name = 'CancelOperationError';
+  }
+}
+
```
xinyiz1226: Do we want to define SystemError class and UserError class and put either of two as base class for all these error classes. For UserError class, we can add one additional field like "suggestion". It will be simpler for any caller catching errors and treat them differently for user/system errors.

src/common/Error/Error.ts
```diff
@@ -0,0 +1,121 @@
...
+  constructor(operation: string, suggestedOperation?: string) {
+    super(`Failed to ${operation} because prerequisite is not met. ${
+        suggestedOperation}`);
+    this.name = 'PrerequisiteNotMetError';
+  }
+}
+
+
+// System Error
+export class OperationFailedError extends Error {
```
xinyiz1226: for this one, does it an user error or system error?

src/Models/ArduinoDeviceBase.ts
dilin-MS: fixed.

src/Models/ArduinoDeviceBase.ts
dilin-MS: fixed.

src/common/Error/Error.ts
dilin-MS: fixed.

PR #896, opened 2019-12-30
Remove project path from telemetry and add hash project name

dilin-MS:


src/utils.ts
```diff
@@ -888,4 +889,17 @@ export function shouldShowLandingPage(context: vscode.ExtensionContext):
...
   return !hasPopUp;
+}
+
+/**
+ * Hash a string and get hash value.
+ * @param stringToHash string to hash
+ * @param algorithm hash algorithm
+ */
+export function getHashFromString(
+    stringToHash: string, algorithm = 'md5'): string {
```
eriolchan: It is better to use sha256 instead of md5

PR #908, opened 2020-01-09
Replace gts with eslint

dilin-MS:
Replace gts with eslint and prettier.

Leave out `views/` `vendor/node-usb-native` and `assets/` folder since javascript files in these folder cause too many eslint error. 
Delaying linting these files for the below three reasons:
1. There are tons of eslint error in these files. Fixing them requires much work.
2. Since changes are so many, it requires careful test. Or it will be error-prone.
3. This part of code is relatively infrequently modified.

We can discuss later whether take time to fix eslint errors in these folders.


.eslintrc.json
```diff
@@ -0,0 +1,42 @@
...
+  ],
+  "rules": {
+    "max-len": ["error", {"code": 120, "ignorePattern": "^import\\s.+\\sfrom\\s.+;$", "ignoreStrings": true}],
+    "no-unused-vars": "off",
+    "@typescript-eslint/no-var-requires": "off",
+    "no-dupe-class-members": "off",
+    "@typescript-eslint/no-empty-function": ["error", { "allow": ["constructors"] }],
+    "object-curly-spacing": ["error", "always"],
+    "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/no-use-before-define": "off"
```
eriolchan: why turn off rule "no-use-before-define"?

.vscode/settings.json
```diff
@@ -6,5 +6,25 @@
...
+    },
+    "[javascriptreact]":  {
+    "editor.formatOnSave":  false,
+    },
+    "[typescript]":  {
+    "editor.formatOnSave":  false,
+    },
+    "[typescriptreact]":  {
+    "editor.formatOnSave":  false,
     }
```
eriolchan: why not format on save for TypeScript

azure-pipelines.yml
```diff
@@ -36,8 +36,8 @@ steps:
     inputs:
       verbose: false
 
-  - script: node node_modules/gts/build/src/cli.js check
-    displayName: 'check gts error'
+  - script: npm run check
+    displayName: 'Run check eslint'
```
eriolchan: "Run eslint check" would be better

package.json
```diff
@@ -229,9 +229,8 @@
     "watch": "tsc -watch -p ./ && node scripts/copyVendor.js",
     "postinstall": "node ./node_modules/vscode/bin/install",
     "test": "npm run compile && node ./node_modules/vscode/bin/test",
-    "check": "gts check",
-    "clean": "gts clean",
-    "fix": "gts fix",
+    "check": "./node_modules/eslint/bin/eslint.js . --ext .js,.jsx,.ts,.tsx",
```
eriolchan: Does eslint provide command as something like eslint instead of directly call bin/eslint?

package.json
```diff
@@ -229,9 +229,8 @@
...
+    "fix": "./node_modules/eslint/bin/eslint.js . --ext .js,.jsx,.ts,.tsx --fix",
```
eriolchan: can we merge check and fix into one script since they are always called at the same time? name the script as "eslint"?

src/AzureOperator.ts
```diff
@@ -1,42 +1,50 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-'use strict';
+"use strict";
```
eriolchan: we can get rid of declaration of "use strict" if we set "strict: true" in tsconfig.json.

src/DigitalTwin/pnp/src/api/apiProvider.ts
```diff
@@ -10,21 +10,32 @@ import { UIConstants } from "../view/uiConstants";
  * Api provider for extension integration
  */
 export class ApiProvider {
-  constructor(private readonly modelRepositoryManager: ModelRepositoryManager) {}
+  constructor(
```
eriolchan: it looks like the max length = 120 doesn't take effect since code in pnp folder is format by tslint with 120 length

src/DigitalTwin/pnp/src/common/utility.ts
```diff
@@ -36,11 +42,15 @@ export class Utility {
...
+  static replaceAll(
+    str: string,
+    replacement: Map<string, string>,
+    caseInsensitive = false
+  ): string {
     const flag = caseInsensitive ? "ig" : "g";
     const keys = Array.from(replacement.keys());
     const pattern = new RegExp(keys.join("|"), flag);
-    return str.replace(pattern, (matched) => {
+    return str.replace(pattern, matched => {
```
eriolchan: here is different from tslint of rule "arrow-parens"

src/DigitalTwin/pnp/src/common/utility.ts
```diff
@@ -94,44 +114,61 @@ export class Utility {
    * @param modelId model id
    * @param content model content
    */
-  // tslint:disable-next-line:no-any
-  static async createModelFile(folder: string, modelId: string, content: any): Promise<void> {
-    const type: ModelType = DeviceModelManager.convertToModelType(content[DigitalTwinConstants.TYPE]);
+  static async createModelFile(
+    folder: string,
+    modelId: string,
+    // eslint-disable-next-line  @typescript-eslint/no-explicit-any
```
eriolchan: do we need "no-explicit-any" comment? tslint allows type of any

.eslintrc.json
```diff
@@ -0,0 +1,42 @@
...
+    "amd": true,
+    "commonjs": true
+  },
+  "plugins": [
+    "@typescript-eslint/eslint-plugin"
+  ],
+  "rules": {
+    "max-len": ["error", {"code": 120, "ignorePattern": "^import\\s.+\\sfrom\\s.+;$", "ignoreStrings": true}],
+    "no-unused-vars": "off",
+    "@typescript-eslint/no-var-requires": "off",
```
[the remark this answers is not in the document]
dilin-MS: Fixed.

.eslintrc.json
```diff
@@ -0,0 +1,42 @@
...
+    "no-dupe-class-members": "off",
```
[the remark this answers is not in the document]
dilin-MS: Fixed.

.eslintrc.json
dilin-MS: Before, this error occurs a lot in js file. I change the lint principle: only linting .ts files. And this "no-use-before-define" rule can be removed here.
Fixed.

.vscode/settings.json
dilin-MS: "formatOnSave" is conflict with "autoFixOnSave", so we should set autoFixOnSave" as true and "formatOnSave" as false. See https://github.com/prettier/prettier-vscode/issues/315

azure-pipelines.yml
dilin-MS: Fxied.

package.json
dilin-MS: Use eslint from node_modules path allow us to use it without install eslint globally.
Outside user trigger `npm run check` to run eslint check and `npm run fix` to run eslint auto fix.

package.json
dilin-MS: The two script do different things. If you just want to check you can run `npm run check` which will execute eslint check. We use this command in CI build.
If you would like to auto fix the eslint error while checking them, run the `npm run fix`. Developers can leverage this command to auto-fix eslint error while developing.

src/DigitalTwin/pnp/src/api/apiProvider.ts
dilin-MS: Fixed.

src/DigitalTwin/pnp/src/common/utility.ts
dilin-MS: Yes. eslint does not allow any.

src/DigitalTwin/pnp/src/common/utility.ts
dilin-MS: From https://eslint.org/docs/rules/arrow-parens, it seems no use of parens when they are not required is recommended.

src/AzureOperator.ts
dilin-MS: Fixed.

Issue #730, opened 2019-08-20
IoT Hub connection details are stored in the workspace file


jimbobbennett:
If I create a new project containing an IoT Hub, the connection details including the IoT Hub connection string are embedded in the workspace file.

This is dangerous, as I'd want to check these files into source code control, which is a problem for open source code, or developers using different hubs.

These connection details are also in the `azureconfig.json` file, so could this be added to the .gitignore to avoid checking in credentials?

shizn:
@ArthurMa1978 Could you help take care of this issue?

JerryYangKai:
Hi @jimbobbennett , thanks for your suggestion. We will generate a .gitignore containing these files when user create a new project. This will be fixed in next feature released.

jimbobbennett:
@JerryYangKai - yes for the azureconfig.json. But for the workspace file it needs to be checked in, so not in the .gitignore. For this can you remove the connection details from the workspace files.

chagong:
@jimbobbennett This issue has been fixed by moving credential info to .azureconfig.json file. The workspace file will not contain any sensitive info any more. Please check it out.

PR #912, opened 2020-01-21
Save credential info into .azureconfig.json instead of workspace file

chagong:
Issue link:
[https://github.com/microsoft/vscode-iot-workbench/issues/730]

Major changes:

- Remove writing azure connection info into .code-workspace file;
- Change IoTHub, IoTHubDevice, AzureFuntions modules to pass on azureconfig file handler info;
- Change IoTWorkspaceProject accordingly;


chagong:
Change AZ3166Device, Esp32Device, IoTButtonDevice, RaspberryPiDevice to get azure credential info grom azure config file instead of workspace file.

src/Models/IoTHubDevice.ts
```diff
@@ -86,19 +104,50 @@ export class IoTHubDevice implements Component, Provisionable {
...
+    );
+
+    if (!fs.existsSync(azureConfigFilePath)) {
+      return false;
+    }
+
+    let azureConfigs: AzureConfigs;
+
+    try {
+      azureConfigs = JSON.parse(fs.readFileSync(azureConfigFilePath, "utf8"));
```
chagong: Better to use another util to handler JSON file.

src/Models/IoTButtonDevice.ts
```diff
@@ -7,14 +7,14 @@ import * as path from "path";
 import * as request from "request-promise";
```
chagong: Delete this module.

src/Models/Esp32Device.ts
```diff
@@ -62,6 +63,8 @@ export class Esp32Device extends ArduinoDeviceBase {
 
   name = "Esp32Arduino";
 
+  private azureConfigFileHandler: AzureConfigFileHandler;
```
dilin-MS: Consider create AzureComponent class and move this field to the class since this field appears for many comnents whch needs azure configuration.

src/Models/Esp32Device.ts
```diff
@@ -75,6 +78,7 @@ export class Esp32Device extends ArduinoDeviceBase {
     if (templateFiles) {
       this.templateFiles = templateFiles;
     }
+    this.azureConfigFileHandler = new AzureConfigFileHandler(devicePath + "/..");
```
dilin-MS: Same as above and suggest have a variable such as "projectRootPath" to represent `devicePath + "/.."`, which is more humanreadable.

src/Models/AZ3166Device.ts
```diff
@@ -168,7 +172,21 @@ export class AZ3166Device extends ArduinoDeviceBase {
...
+        // check if there are any IOT Hub Device Components
+        for (const componentConfig of componentConfigs) {
+          if (componentConfig.type === ComponentType.IoTHubDevice) {
+            if (componentConfig.componentInfo) {
+              deviceConnectionString = componentConfig.componentInfo.values.iotHubDeviceConnectionString;
+              break;
+            }
+          }
+        }
+      }
```
dilin-MS: This part of logic appears multiple times. Suggest move to a function to both reduce replicated code and increase code clarity.

src/Models/Esp32Device.ts
chagong: Are you suggesting to define a static class to handle AzureConfig info? I've ever considered it. It will bring changes to lots of files. AzureConfigFileHandler class is an existing class for handling azure config file. Define a "azureConfigFileHandler" member in device/module classes is the current practice.

src/Models/AZ3166Device.ts
chagong: Accept. It has better way to do that.

eriolchan: It seem this could be made as a method of AzureConfigFileHandler which returning config value with key under specified component. This would be more reusable than directly manipulate the config file content.

src/Models/AzureFunctions.ts
```diff
@@ -48,7 +42,8 @@ export class AzureFunctions implements Component, Provisionable, Deployable {
   private azureAccountExtension: AzureAccount | undefined = getExtension(ExtensionName.AzureAccount);
   private functionLanguage: string | null;
   private functionFolder: string;
-
+  private projectRootPath: string;
+  private azureConfigFileHandler: AzureConfigFileHandler;
```
eriolchan: I am not sure if we could make AzureConfigFileHandler as a utility instead of always passing it to consturctor?

src/Models/AzureFunctions.ts
```diff
@@ -256,10 +253,21 @@ export class AzureFunctions implements Component, Provisionable, Deployable {
...
+      }
+      if (!componentConfig.componentInfo) {
+        return false;
+      }
+      const iotHubConnectionString = componentConfig.componentInfo.values.iotHubConnectionString;
+      if (!iotHubConnectionString) {
+        throw new Error("Unable to find Iot Hub connection in the project. Please retry Azure Provision.");
+      }
+      const eventHubConnectionString = componentConfig.componentInfo.values.eventHubConnectionString;
+      const eventHubConnectionPath = componentConfig.componentInfo.values.eventHubConnectionPath;
```
eriolchan: It would be better to encapsulate the read/write operation in component level as method of AzureConfigFileHandler utility, and this would be more reusable.

src/Models/IoTHubDevice.ts
```diff
@@ -138,12 +197,23 @@ export class IoTHubDevice implements Component, Provisionable {
...
+      }
+      await this.azureConfigFileHandler.updateComponent(type, iotHubComponentIndex, componentInfo);
+    } else {
+      const newIotHubConfig: AzureComponentConfig = {
+        id: this.id,
+        folder: "",
+        name: "",
+        dependencies: this.dependencies,
+        type: this.componentType,
+        componentInfo
```
eriolchan: what if componentInfo is undefined here? Do we need to check componentInfo first?

src/Models/IoTWorkspaceProject.ts
```diff
@@ -244,18 +241,32 @@ export class IoTWorkspaceProject extends IoTWorkbenchProjectBase {
     await iotHub.load();
     this.componentList.push(iotHub);
 
-    const iothubDevice = new ioTHubDeviceModule.IoTHubDevice(this.channel);
-    this.componentList.push(iothubDevice);
+    const iotHubDevice = new ioTHubDeviceModule.IoTHubDevice(this.projectRootPath, this.channel, [
+      {
+        component: iotHub,
```
eriolchan: if we need this kind of structure, we could define it as interface

src/Models/AZ3166Device.ts
```diff
@@ -168,7 +172,21 @@ export class AZ3166Device extends ArduinoDeviceBase {
       return retValue;
     } else if (configSelection.detail === "Config Connection String") {
       // Get IoT Hub device connection string from config
-      let deviceConnectionString = ConfigHandler.get<string>(ConfigKey.iotHubDeviceConnectionString);
+      const componentConfigs = await this.azureConfigFileHandler.getSortedComponents(ScaffoldType.Workspace);
```
eriolchan: What is the purpose of getSortedComponents method? Since I didn't seen any code change about AzureConfigFileHandler, I assume you just use the current implementation of that class. You could check if the current AzureConfigFileHandler methods are well designed and tailor it in a proper way.

src/Models/Esp32Device.ts
chagong: Accept.

src/Models/AZ3166Device.ts
chagong: Added a new method in AzureConfigFileHandler named "getComponentByTYpe".

src/Models/IoTHubDevice.ts
chagong: It's allowed that if componentInfo is undefined. "appendComponent" function takes component as param, it's good as long as component is not undefined.

src/Models/IoTWorkspaceProject.ts
chagong: "IoTHubDevice" implements the interfaces "Component" and "Provisionable". Inside "Provisionable" it already has a member "dependencies". The old code didn't pass the dependencies info when creating IoTHubDevice instance. In fact IoTHubDevice should have "IoTHub" as dependency.

src/Models/AZ3166Device.ts
chagong: Already changed to use another new-added method to get componentConfig.

src/Models/AzureFunctions.ts
chagong: I ever considered this way, this is likely to make "AzureConfigFileHandler" as a static class. I was not so sure if all our projects would need Azure Config support. So I just leave it as it is.

src/Models/AzureFunctions.ts
```diff
@@ -258,16 +237,37 @@ export class AzureFunctions implements Component, Provisionable, Deployable {
...
+      );
+    }
+    const eventHubConnectionString = componentConfig.componentInfo.values.eventHubConnectionString;
+    const eventHubConnectionPath = componentConfig.componentInfo.values.eventHubConnectionPath;
 
     if (!eventHubConnectionString) {
-      throw new WorkspaceConfigNotFoundError(ConfigKey.eventHubConnectionString);
+      throw new AzureConfigNotFoundError(
+        `eventHubConnectionString of config id ${iotHubId}`,
+        "Please provision Azure Services"
```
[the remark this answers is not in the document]
chagong: OK.

src/Models/AzureFunctions.ts
```diff
@@ -341,19 +357,12 @@ export class AzureFunctions implements Component, Provisionable, Deployable {
...
-    let azureConfigs: AzureConfigs = { componentConfigs: [] };
-    azureConfigs = await AzureConfigFileHandler.loadAzureConfigs(type, azureConfigFilePath);
-
-    const azureFunctionsConfig = azureConfigs.componentConfigs.find(config => config.id === this.id);
-    if (azureFunctionsConfig) {
-      // TODO: update the existing setting for the provision result
+    const componentIndex = await this.azureConfigFileHandler.getComponentIndexById(type, this.id);
+    if (componentIndex > -1) {
+      if (!componentInfo) {
+        throw new ArgumentEmptyOrNullError("AzureFunctions updateConfigSettings", `componentInfo.`);
```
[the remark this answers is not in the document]
chagong: OK.

src/Models/CosmosDB.ts
```diff
@@ -143,7 +138,7 @@ export class CosmosDB implements Component, Provisionable {
       channelPrintJsonObject(this.channel, cosmosDBDeploy);
 
       for (const dependency of this.dependencies) {
-        const componentConfig = await this.azureConfigHandler.getComponentById(scaffoldType, dependency.id);
+        const componentConfig = await this.azureConfigFileHandler.getComponentById(scaffoldType, dependency.id);
         if (!componentConfig) {
           throw new OperationFailedError(`find component with id ${dependency.id}`, "", "");
```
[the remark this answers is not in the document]
chagong: Should not be a user error. This is from error-handling merge. I will change it to AzureConfigNotFoundError.

src/Models/Esp32Device.ts
```diff
@@ -123,11 +128,19 @@ export class Esp32Device extends ArduinoDeviceBase {
...
+        ScaffoldType.Workspace,
+        ComponentType.IoTHubDevice
+      );
+      if (componentConfig) {
+        deviceConnectionString = componentConfig.componentInfo?.values.iotHubDeviceConnectionString;
+      }
 
       if (!deviceConnectionString) {
-        throw new WorkspaceConfigNotFoundError(ConfigKey.iotHubDeviceConnectionString);
+        throw new AzureConfigNotFoundError("iotHubDeviceConnectionString", "Please provision Azure Servcies");
```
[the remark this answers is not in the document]
chagong: Good catch!

src/Models/StreamAnalyticsJob.ts
```diff
@@ -257,7 +253,7 @@ export class StreamAnalyticsJob implements Component, Provisionable, Deployable
     }
 
     for (const dependency of this.dependencies) {
-      const componentConfig = await this.azureConfigHandler.getComponentById(scaffoldType, dependency.id);
+      const componentConfig = await this.azureConfigFileHandler.getComponentById(scaffoldType, dependency.id);
       if (!componentConfig) {
         throw new OperationFailedError(
           "provision stream analystics job",
```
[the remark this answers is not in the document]
chagong: Should not be a user error. This is from error-handling merge. I will change it to AzureConfigNotFoundError.

PR #920, opened 2020-02-18
fix: pnp context value support both array and string

eriolchan:
make @context support both array and string for PnP IntelliSense

### DIFF
--- a/src/DigitalTwin/pnp/src/intelliSense/intelliSenseUtility.ts
+++ b/src/DigitalTwin/pnp/src/intelliSense/intelliSenseUtility.ts
@@ -88,9 +88,15 @@ export class IntelliSenseUtility {
    * @param node json node
    */
   static isDigitalTwinContext(node: parser.Node): boolean {
-    // assume @context is string node
+    // @context accept both array and string
     if (node.type === JsonNodeType.String) {
-      return DigitalTwinConstants.CONTEXT_REGEX.test(node.value as string);
+      return (node.value as string) === DigitalTwinConstants.CONTEXT_TEMPLATE;
+    } else if (node.type === JsonNodeType.Array && node.children) {
+      for (const child of node.children) {
+        if (child.type === JsonNodeType.String && (child.value as string) === DigitalTwinConstants.CONTEXT_TEMPLATE) {
+          return true;
+        }
+      }
     }
     return false;
   }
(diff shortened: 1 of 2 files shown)


src/DigitalTwin/pnp/src/intelliSense/intelliSenseUtility.ts
```diff
@@ -88,9 +88,15 @@ export class IntelliSenseUtility {
    * @param node json node
    */
   static isDigitalTwinContext(node: parser.Node): boolean {
-    // assume @context is string node
+    // @context accept both array and string
```
dooriya: Seems this is for `@context` node diagnostic. 
How about the IntelliSense support for `@context` property for both single value and multiple values?

As I remember, currently we use the `@contex`t property to determine whether a JSON file is a DTDL file so that we can apply the DTDL IntelliSense on it.  And we don't expect users to modify the `@context` property. 

In DTDL v2, user can reference multiple context files within the `@context` property.  We might need to update the defaultcapabilityModel/interface sample file to use the array format rather than the string format. User then can add another context file (e.g. his own extension context) into this property.
`"@context": "http://azureiot.com/v1/contexts/IoTModel.json"` is changed to `"@context": ["http://azureiot.com/v1/contexts/IoTModel.json"]`. 

See [this link](https://json-ld.org/spec/latest/json-ld/#lists-and-sets) for more details on the JSON-LD multi-values property.

eriolchan: I prefer this fix is still for DTDL v1. That is, @context supports array type (which is exported by IoT Central) in diagnostics, and we suggest user to use string type in auto-completion (because there is only one @context value in DTDL v1).

You are right that when we try to support DTDL v2, we may think about if using array for @context in template file. Since we haven't got enough real sample of DTDL v2 which proves v2 suggest using multiple value for @context, I prefer to leave this change later for the task of supporting v2.

dooriya: Good to me if this is for v1 fix.

---
In reply to: [380465715](https://github.com/microsoft/vscode-iot-workbench/pull/920#discussion_r380465715) [](ancestors = 380465715)

PR #924, opened 2020-02-20
Reduce duplicate code of getting dependent extension

dilin-MS:


src/Models/Apis.ts
```diff
@@ -4,22 +4,25 @@
...
-import { AzureAccountCommands } from "../common/Commands";
+import { AzureAccountCommands, VscodeCommands } from "../common/Commands";
 import { DependentExtensionNotFoundError } from "../common/Error/OperationFailedErrors/DependentExtensionNotFoundError";
-
-import { ExtensionName } from "./Interfaces/Api";
+import { ExtensionName, DependentExtensions, ExtensionNameIdMap } from "./Interfaces/Api";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export function getExtension(name: ExtensionName): any {
+  // }
```
eriolchan: why the commented code here?
```
// }
```

src/Models/Apis.ts
```diff
@@ -4,22 +4,25 @@
...
   switch (name) {
-    case ExtensionName.Toolkit: {
-      const toolkit = vscode.extensions.getExtension("vsciot-vscode.azure-iot-toolkit");
-      return toolkit ? toolkit.exports : undefined;
-    }
     case ExtensionName.AzureAccount: {
-      const azureAccount = vscode.extensions.getExtension<AzureAccount>("ms-vscode.azure-account");
+      const azureAccount = vscode.extensions.getExtension<AzureAccount>(DependentExtensions.azureAccount);
```
eriolchan: could AzureAccount behavior the same as other extensions?

src/Models/Apis.ts
```diff
@@ -39,3 +42,22 @@ export async function checkAzureLogin(): Promise<boolean> {
...
+  if (!getExtension(extensionName)) {
+    const choice = await vscode.window.showInformationMessage(
+      `${extensionName} extension is required for the current project. Do you want to install it from marketplace?`,
+      "Yes",
+      "No"
+    );
+    if (choice === "Yes") {
+      vscode.commands.executeCommand(
+        VscodeCommands.VscodeOpen,
+        vscode.Uri.parse(("vscode:extension/" + ExtensionNameIdMap[ExtensionName.Arduino]) as string)
```
eriolchan: I think it is extensionName instead of ExtensionName.Arduino here?

src/Models/Interfaces/Api.ts
```diff
@@ -3,5 +3,24 @@
...
+
+export class DependentExtensions {
+  static readonly azureFunctions = "ms-azuretools.vscode-azurefunctions";
+  static readonly arduino = "vsciot-vscode.vscode-arduino";
+  static readonly remote = "ms-vscode-remote.vscode-remote-extensionpack";
+  static readonly toolkit = "vsciot-vscode.azure-iot-toolkit";
+  static readonly azureAccount = "ms-vscode.azure-account";
+}
+
+export const ExtensionNameIdMap: Record<ExtensionName, DependentExtensions> = {
```
eriolchan: why prefer Record<> to Map?

src/Models/Apis.ts
dilin-MS: Fixed.

src/Models/Apis.ts
dilin-MS: Fixed.

src/Models/Apis.ts
dilin-MS: I tested and it is okay to remove the type defined for AzureAccout. Removed the type restriction.

src/Models/Interfaces/Api.ts
dilin-MS: Map and Record both work here.
For the difference, Map<K, V> does not require all members of type K to appear in the map, while Record<K, V> requires all members of K to appear in the record.
I change record to map here, since we don't have a strong restriction to ask all members of ExtensionName to appear in the map.

> Record<K, V> can be used to map the properties of a type to another type.

[Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkt)

src/Models/Apis.ts
```diff
@@ -3,30 +3,27 @@
...
-      const extension = vscode.extensions.getExtension(ExtensionNameIdMap[name] as string);
-      return extension ? extension.exports : undefined;
-    }
-    default: {
-      return undefined;
+  const extensionId = ExtensionNameIdMap.get(name);
+  if (extensionId) {
+    const extension = vscode.extensions.getExtension(extensionId);
+    if (extension) {
+      switch (name) {
```
eriolchan: missing ExtensionName.Remote here?

src/Models/Apis.ts
```diff
@@ -3,30 +3,27 @@
...
-    case ExtensionName.Toolkit:
-    case ExtensionName.Remote:
-    case ExtensionName.AzureFunctions:
-    case ExtensionName.Arduino: {
-      const extension = vscode.extensions.getExtension(ExtensionNameIdMap[name] as string);
-      return extension ? extension.exports : undefined;
-    }
-    default: {
-      return undefined;
+  const extensionId = ExtensionNameIdMap.get(name);
```
eriolchan: Since ExtensionName is an enum, can we just put extensionId as the value of it instead of manually setup mapping from extensionName to extensinId?

src/Models/Apis.ts
```diff
@@ -3,27 +3,27 @@
...
-    case ExtensionName.AzureAccount: {
-      const azureAccount = vscode.extensions.getExtension<AzureAccount>("ms-vscode.azure-account");
-      return azureAccount ? azureAccount.exports : undefined;
-    }
-    default: {
-      return undefined;
+  const extensionId = ExtensionNameIdMap.get(name);
+  if (extensionId) {
+    const extension = vscode.extensions.getExtension(extensionId);
+    if (extension) {
```
eriolchan: It seems that it is not necessary to list all enum, since they the same.
``` return extension && entension.exports ? extension.exports : undefined;
```

src/Models/Apis.ts
dilin-MS: “Remote” and "Arduino" follows the default behavior, return extension directly instead of return extension.exports since their exports are undefined.

src/Models/Apis.ts
dilin-MS: ExtensionName value will be used in other scenario, mainly in messages for user to install from marketplace. Hence, we cannot directly use extensionId as the value of ExtensionName and discard the name string.

PR #925, opened 2020-02-20
Fix foreach bug

dilin-MS:
In TypeScript, usage of `foreach` should be careful. If function inside `foreach` is asynchronous, `foreach` will not wait until all items are finished. (See [Reference](https://lavrton.com/javascript-loops-how-to-handle-async-await-6252dd3c795/)).

this PR:
1. Fix foreach with async functions. Change to `Promise.all()`. Fix the [bug](https://dev.azure.com/mseng/VSIoT/_workitems/edit/1679203).
2. Change `checkPrerequisite()` function to return `Promise<void>` instead of `Promise<boolean>`. User will get error message when operation failed for not finding dependent extensions.

### DIFF
--- a/src/Models/ArduinoDeviceBase.ts
+++ b/src/Models/ArduinoDeviceBase.ts
@@ -23,6 +23,7 @@ import { OTA } from "./OTA";
 import { DirectoryNotFoundError } from "../common/Error/OperationFailedErrors/DirectoryNotFoundError";
 import { FileNotFoundError } from "../common/Error/OperationFailedErrors/FileNotFound";
 import { checkExtensionAvailable } from "./Apis";
+import { DependentExtensionNotFoundError } from "../common/Error/OperationFailedErrors/DependentExtensionNotFoundError";
 import { ExtensionName } from "./Interfaces/Api";
 
 const constants = {
@@ -84,8 +85,11 @@ export abstract class ArduinoDeviceBase implements Device {
     return await checkExtensionAvailable(ExtensionName.Arduino);
   }
 
-  async checkPrerequisites(): Promise<boolean> {
-    return await ArduinoDeviceBase.isAvailable();
+  async checkPrerequisites(operation: string): Promise<void> {
+    const res = await ArduinoDeviceBase.isAvailable();
+    if (!res) {
+      throw new DependentExtensionNotFoundError(operation, ExtensionName.Arduino);
+    }
   }
 
   async compile(): Promise<boolean> {
--- a/src/Models/ContainerDeviceBase.ts
+++ b/src/Models/ContainerDeviceBase.ts
@@ -66,8 +66,8 @@ export abstract class ContainerDeviceBase implements Device {
     return this.componentType;
   }
 
-  async checkPrerequisites(): Promise<boolean> {
-    return true;
+  async checkPrerequisites(): Promise<void> {
+    // Do nothing.
   }
 
   async load(): Promise<void> {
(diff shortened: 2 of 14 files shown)


src/Models/IoTContainerizedProject.ts
```diff
@@ -105,16 +104,9 @@ export class IoTContainerizedProject extends IoTWorkbenchProjectBase {
...
-      const res = await item.checkPrerequisites();
-      if (!res) {
-        throw new OperationFailedError(
-          "create component for iot containerized project",
-          "component prerequisites are not met.",
-          "Please check out error message in the output window."
-        );
-      }
-    });
+    for (const item of this.componentList) {
```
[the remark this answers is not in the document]

src/Models/RemoteExtension.ts
```diff
@@ -25,12 +25,10 @@ export class RemoteExtension {
     return await checkExtensionAvailable(ExtensionName.Remote);
   }
 
-  static async checkRemoteExtension(): Promise<void> {
+  static async checkRemoteExtension(operation: string): Promise<void> {
```
dooriya: nit: can we manage the "operation" as enum type here?
It can provide us an overview of what kind of operations will need to check remote extension. Another benefit to use enum type here is improve the re-usability for the same operation used in multiple places.

src/Models/ArduinoDeviceBase.ts
```diff
@@ -84,8 +85,11 @@ export abstract class ArduinoDeviceBase implements Device {
     return await checkExtensionAvailable(ExtensionName.Arduino);
   }
 
-  async checkPrerequisites(): Promise<boolean> {
-    return await ArduinoDeviceBase.isAvailable();
+  async checkPrerequisites(operation: string): Promise<void> {
```
dooriya: Same here as I comment in the `checkRemoteExtension` api.

src/Models/IoTContainerizedProject.ts
```diff
@@ -76,7 +75,7 @@ export class IoTContainerizedProject extends IoTWorkbenchProjectBase {
     openInNewWindow: boolean
   ): Promise<void> {
     // Can only create project locally
-    await RemoteExtension.checkRemoteExtension();
+    await RemoteExtension.checkRemoteExtension("create iot containerized project");
```
dooriya: The same operation string appears in many places.

src/Models/ContainerDeviceBase.ts
```diff
@@ -66,8 +66,8 @@ export abstract class ContainerDeviceBase implements Device {
     return this.componentType;
   }
 
-  async checkPrerequisites(): Promise<boolean> {
-    return true;
+  async checkPrerequisites(): Promise<void> {
```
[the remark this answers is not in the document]
dilin-MS: Currently `checkPrerequisites(operation: string): Promise<void>;` is the required method of interface `Component`.
Both `ContainerDeviceBase` and `ArduinoDeviceBase` needs to implement this method.
Is it necessary to move `checkPrerequisite()` to be another interface like `ExtensionDependent`?

For the different method signature:

1. ts allows ignoring extra function parameters. More info see [link](https://stackoverflow.com/questions/51727721/typescript-interface-implementing-doesnt-check-method-parameters). `parameter operation: string` does appears because they are not needed in the implementation of the function. 
2. `eslint` does not allow unused vars.

src/Models/RemoteExtension.ts
dilin-MS: The `Operation: string` is used when in error message to help know what operation throws error. The value of operation varies a lot through the whole project. This is why I didn't set it as enum at the very beginning.

If we have a enum type for all operations, the list will be very large. 
E.g. "create function application", "get credential from subscription id", "crete iot containerized project", "open project in remote", "configure project environment", "generate CRC", etc. 
Really a big list since operations varies.

src/Models/ArduinoDeviceBase.ts
dilin-MS: See above.

src/Models/IoTContainerizedProject.ts
dilin-MS: See above.

PR #928, opened 2020-02-24
Fix Arduino project lazy creation

dilin-MS:
Fix [bug](https://dev.azure.com/mseng/VSIoT/_workitems/edit/1679307):

When creating iot workspace project, steps sequence should be:
init components -> check prerequisites(will check dependent extensions) -> create componens.

This PR move all files creation after prerequisites are checked and satisfied. Thus project will NOT be created if prerequistes are not satisfied.

### DIFF
--- a/src/Models/AzureFunctions.ts
+++ b/src/Models/AzureFunctions.ts
@@ -334,10 +334,9 @@ export class AzureFunctions implements Component, Provisionable, Deployable {
   async updateConfigSettings(type: ScaffoldType, componentInfo?: ComponentInfo): Promise<void> {
     const componentIndex = await this.azureConfigFileHandler.getComponentIndexById(type, this.id);
     if (componentIndex > -1) {
-      if (!componentInfo) {
-        throw new ArgumentEmptyOrNullError("AzureFunctions updateConfigSettings", `componentInfo`);
+      if (componentInfo) {
+        await this.azureConfigFileHandler.updateComponent(type, componentIndex, componentInfo);
       }
-      await this.azureConfigFileHandler.updateComponent(type, componentIndex, componentInfo);
     } else {
       const newAzureFunctionsConfig: AzureComponentConfig = {
         id: this.id,
--- a/src/Models/AzureFunctions.ts
+++ b/src/Models/AzureFunctions.ts
@@ -334,9 +334,10 @@ export class AzureFunctions implements Component, Provisionable, Deployable {
   async updateConfigSettings(type: ScaffoldType, componentInfo?: ComponentInfo): Promise<void> {
     const componentIndex = await this.azureConfigFileHandler.getComponentIndexById(type, this.id);
     if (componentIndex > -1) {
-      if (componentInfo) {
-        await this.azureConfigFileHandler.updateComponent(type, componentIndex, componentInfo);
+      if (!componentInfo) {
+        throw new ArgumentEmptyOrNullError("update config settings of IoTHub", "component info");
       }
+      await this.azureConfigFileHandler.updateComponent(type, componentIndex, componentInfo);
     } else {
       const newAzureFunctionsConfig: AzureComponentConfig = {
         id: this.id,
--- a/src/Models/IoTHub.ts
+++ b/src/Models/IoTHub.ts
@@ -15,7 +15,6 @@ import { AzureUtility } from "./AzureUtility";
 import { ExtensionName } from "./Interfaces/Api";
 import { Component, ComponentType } from "./Interfaces/Component";
 import { Provisionable } from "./Interfaces/Provisionable";
-import { ArgumentEmptyOrNullError } from "../common/Error/OperationFailedErrors/ArgumentEmptyOrNullError";
 
 export class IoTHub implements Component, Provisionable {
   dependencies: DependencyConfig[] = [];
@@ -155,10 +154,9 @@ export class IoTHub implements Component, Provisionable {
     const iotHubComponentIndex = await this.azureConfigFileHandler.getComponentIndexById(type, this.id);
 
     if (iotHubComponentIndex > -1) {
-      if (!componentInfo) {
-        throw new ArgumentEmptyOrNullError("IoTHub updateConfigSettings", "componentInfo");
+      if (componentInfo) {
+        await this.azureConfigFileHandler.updateComponent(type, iotHubComponentIndex, componentInfo);
       }
-      await this.azureConfigFileHandler.updateComponent(type, iotHubComponentIndex, componentInfo);
     } else {
       const newIoTHubConfig: AzureComponentConfig = {
         id: this.id,
--- a/src/Models/IoTHub.ts
+++ b/src/Models/IoTHub.ts
@@ -15,6 +15,7 @@ import { AzureUtility } from "./AzureUtility";
 import { ExtensionName } from "./Interfaces/Api";
 import { Component, ComponentType } from "./Interfaces/Component";
 import { Provisionable } from "./Interfaces/Provisionable";
+import { ArgumentEmptyOrNullError } from "../common/Error/OperationFailedErrors/ArgumentEmptyOrNullError";
 
 export class IoTHub implements Component, Provisionable {
   dependencies: DependencyConfig[] = [];
@@ -154,9 +155,10 @@ export class IoTHub implements Component, Provisionable {
     const iotHubComponentIndex = await this.azureConfigFileHandler.getComponentIndexById(type, this.id);
 
     if (iotHubComponentIndex > -1) {
-      if (componentInfo) {
-        await this.azureConfigFileHandler.updateComponent(type, iotHubComponentIndex, componentInfo);
+      if (!componentInfo) {
+        throw new ArgumentEmptyOrNullError("update config settings of azure functions", "component info");
       }
+      await this.azureConfigFileHandler.updateComponent(type, iotHubComponentIndex, componentInfo);
     } else {
       const newIoTHubConfig: AzureComponentConfig = {
         id: this.id,
--- a/src/Models/IoTWorkspaceProject.ts
+++ b/src/Models/IoTWorkspaceProject.ts
@@ -106,7 +106,7 @@ export class IoTWorkspaceProject extends IoTWorkbenchProjectBase {
     );
 
     // Init and load azure components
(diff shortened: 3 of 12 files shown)


src/Models/IoTWorkspaceProject.ts
```diff
@@ -114,56 +123,50 @@ export class IoTWorkspaceProject extends IoTWorkbenchProjectBase {
...
+    );
+    await FileUtility.writeJsonFile(createTimeScaffoldType, this.workspaceConfigFilePath, workspace);
+
     // Create components
     try {
-      for (let i = 0; i < this.componentList.length; i++) {
-        await this.componentList[i].create();
-      }
+      await Promise.all(
+        this.componentList.map(async component => {
```
chagong: Nice change.

src/Models/IoTWorkspaceProject.ts
```diff
@@ -359,35 +346,38 @@ export class IoTWorkspaceProject extends IoTWorkbenchProjectBase {
...
 
   /**
-   * Init Azure components and azure configs.
+   * Load project config and init Azure components.
    * @param scaffoldType scaffold type
    */
-  private async initAzureConfig(scaffoldType: ScaffoldType): Promise<void> {
-    this.validateProjectRootPath("initialize azure configurations", scaffoldType);
-
+  private async loadAzureConfigAndInitAzureComponents(scaffoldType: ScaffoldType): Promise<void> {
```
eriolchan: What about simplify the name to "loadAzureConfigAndComponents"?

src/Models/IoTWorkspaceProject.ts
```diff
@@ -290,30 +287,22 @@ export class IoTWorkspaceProject extends IoTWorkbenchProjectBase {
...
    * @param componentConfigs azure component configs
    */
-  private async initAzureComponentsWithConfig(
-    scaffoldType: ScaffoldType,
-    componentConfigs: AzureComponentConfig[]
-  ): Promise<void> {
-    this.validateProjectRootPath("initialize azure components with configurations", scaffoldType);
-
-    const components: { [key: string]: Component } = {};
+  private async loadAzureComponentsWithConfig(componentConfigs: AzureComponentConfig[]): Promise<void> {
```
eriolchan: "loadAzureComponentsByConfig" would be a better name?

src/Models/AzureFunctions.ts
```diff
@@ -334,10 +334,9 @@ export class AzureFunctions implements Component, Provisionable, Deployable {
   async updateConfigSettings(type: ScaffoldType, componentInfo?: ComponentInfo): Promise<void> {
     const componentIndex = await this.azureConfigFileHandler.getComponentIndexById(type, this.id);
     if (componentIndex > -1) {
-      if (!componentInfo) {
-        throw new ArgumentEmptyOrNullError("AzureFunctions updateConfigSettings", `componentInfo`);
+      if (componentInfo) {
+        await this.azureConfigFileHandler.updateComponent(type, componentIndex, componentInfo);
```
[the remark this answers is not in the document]
dilin-MS: Changed back to original logic.

src/Models/IoTHub.ts
```diff
@@ -155,10 +154,9 @@ export class IoTHub implements Component, Provisionable {
     const iotHubComponentIndex = await this.azureConfigFileHandler.getComponentIndexById(type, this.id);
 
     if (iotHubComponentIndex > -1) {
-      if (!componentInfo) {
-        throw new ArgumentEmptyOrNullError("IoTHub updateConfigSettings", "componentInfo");
+      if (componentInfo) {
+        await this.azureConfigFileHandler.updateComponent(type, iotHubComponentIndex, componentInfo);
```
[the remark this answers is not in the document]
dilin-MS: Same as above.

src/utils.ts
```diff
@@ -141,16 +142,31 @@ export interface FolderQuickPickItem<T = undefined> extends vscode.QuickPickItem
...
     return "";
   }
+
+  // Try get the "Device" folder
+  const devicePath = ConfigHandler.get<string>(ConfigKey.devicePath);
+  if (devicePath) {
+    const deviceFolder = vscode.workspace.workspaceFolders.find(
+      folder => path.basename(folder.uri.fsPath) === devicePath
+    );
+    if (deviceFolder) {
```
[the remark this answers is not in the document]
dilin-MS: Yes. 
This is a "try to get device folder" function. The logical device folder for iot workspace project is the `/Device` folder, the logical device folder for iot containerized project is the root folder.

 * For iot workspace project, the "Device" folder will be returned.
 * For iot containerized project, the root path (aka the first folder), will be returned.
 * For non-workbench project, we can just return the first folder.

src/Models/IoTWorkspaceProject.ts
dilin-MS: Fixed.

src/Models/IoTWorkspaceProject.ts
dilin-MS: Fixed.

src/Models/IoTWorkspaceProject.ts
```diff
@@ -174,7 +177,7 @@ export class IoTWorkspaceProject extends IoTWorkbenchProjectBase {
   }
 
   async openProject(scaffoldType: ScaffoldType, openInNewWindow: boolean, openScenario: OpenScenario): Promise<void> {
-    this.loadAndInitWorkspaceConfigFilePath(scaffoldType);
+    this.loadWorkspaceConfigFilePath(scaffoldType);
```
dooriya: please "await" this async operation

src/Models/IoTWorkspaceProject.ts
```diff
@@ -203,19 +206,15 @@ export class IoTWorkspaceProject extends IoTWorkbenchProjectBase {
...
-  private async initDevice(
-    boardId: string,
-    scaffoldType: ScaffoldType,
-    templateFilesInfo?: TemplateFileInfo[]
-  ): Promise<void> {
-    if (!(await FileUtility.directoryExists(scaffoldType, this.deviceRootPath))) {
-      throw new DirectoryNotFoundError(
+  private initDeviceComponents(boardId: string, templateFilesInfo?: TemplateFileInfo[]): void {
+    if (!this.deviceRootPath) {
+      throw new ArgumentEmptyOrNullError(
```
dooriya: This is a little bit weird here to check a dependent class member "deviceRootPath" and throw error if it is empty or null. I would suggest putting such required value in the function parameter. Or ensure the value of `deviceRootPath` in the class constructor and no more check its value in this private function.

src/Models/IoTWorkspaceProject.ts
```diff
@@ -359,35 +346,38 @@ export class IoTWorkspaceProject extends IoTWorkbenchProjectBase {
...
+  private async loadAzureConfigAndComponents(scaffoldType: ScaffoldType): Promise<void> {
     const azureConfigFileHandler = new azureComponentConfigModule.AzureConfigFileHandler(this.projectRootPath);
     await azureConfigFileHandler.createIfNotExists(scaffoldType);
 
     const componentConfigs = await azureConfigFileHandler.getSortedComponents(scaffoldType);
 
     if (componentConfigs.length === 0) {
       // Support backward compact
-      await this.initAzureComponentsWithoutConfig(scaffoldType);
+      this.initAzureComponentsWithoutConfig(scaffoldType);
```
dooriya: Don't forget to "await"

src/utils.ts
```diff
@@ -141,16 +142,31 @@ export interface FolderQuickPickItem<T = undefined> extends vscode.QuickPickItem
...
+export function getProjectDeviceRootPath(): string {
   if (
     !vscode.workspace.workspaceFolders ||
     vscode.workspace.workspaceFolders.length === 0 ||
     !vscode.workspace.workspaceFolders[0].uri.fsPath
   ) {
     return "";
   }
+
+  // Try get the "Device" folder
```
dooriya: nit: Try get -> Try to get

src/Models/IoTWorkspaceProject.ts
```diff
@@ -359,35 +346,38 @@ export class IoTWorkspaceProject extends IoTWorkbenchProjectBase {
...
+   * Update azure components information in workspace configuration.
+   * @param projectType project template type
+   * @param workspaceConfig workspace configuration
+   */
+  private initAzureComponents(projectType: ProjectTemplateType, workspaceConfig: Workspace): void {
+    if (!this.projectRootPath) {
+      throw new ArgumentEmptyOrNullError(
+        "init azure components",
+        `project root path: ${this.projectRootPath}`,
+        "Please initialize the project first."
```
dooriya: I would prefer formatting the error message and then pass it to the `ArgumentEmptyOrNullError` rather than add complexity to the constructor. This makes the code less readable.

src/Models/IoTWorkspaceProject.ts
dilin-MS: Nice catch.

src/Models/IoTWorkspaceProject.ts
dilin-MS: Fixed.

src/Models/IoTWorkspaceProject.ts
dilin-MS: There is no strict principle on the problem. I would rather follow the below principles:

1. whether validate parameters inside functions:
* Document the methods well, and encourage the callers to pass good/valid data.
* the parameters anyway! - throwing exceptions when the preconditions are not met.(Unless the function is called too many times that checking parameters will cause huge time costs. E.g. Checking parameters when calculate matrix, etc.)

2. Whether passing private members as parameters into private methods:
* If the private parameter is closely logically relevant with the private method, it can serve as a parameter of the function.

> E.g. int area(int length, int width);
	
* If the private members is more of a global state, then don't pass it as parameter.

In our code, I prefer to put `deviceRootPath` as parameter since it is not initialized in the class constructor, but initialized differently from `create()` and `load()` methods. For other used private parameters, like `this.extensionContext`, `this.channel`, `this.telemetryContext`, I would rather keep them **NOT** as parameter but being accessed directly from the private function.

src/utils.ts

src/Models/IoTWorkspaceProject.ts
dilin-MS: Will discuss further and fix in new PR for all error types.

PR #931, opened 2020-02-25
test: add test case for utility of fs read/write

eriolchan:


src/DigitalTwin/pnp/src/common/utility.ts
```diff
@@ -38,12 +38,10 @@ export class Utility {
    * replace all for content
    * @param str string
    * @param replacement replacement
-   * @param caseInsensitive identify if it is case insensitive
    */
-  static replaceAll(str: string, replacement: Map<string, string>, caseInsensitive = false): string {
-    const flag = caseInsensitive ? "ig" : "g";
```
dilin-MS: Why remove the caseInsensitive parameter?

test/digitalTwin/utility.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  replacement.set("{foo}", "bar");
+
+  test("create file from template", async () => {
+    const templatePath = "templatePath";
+    const filePath = "filePath";
+    const writeOptions = {
+      spaces: Constants.JSON_SPACE,
+      encoding: Constants.UTF8
+    };
+    const template = '{"@id": "{DigitalTwinIdentifier}"}';
```
dilin-MS: >DigitalTwinIdentifier [](start = 32, length = 21)

Can we replace this with Constants.DIGITAL_TWIN_ID_PLACEHOLDER?

test/digitalTwin/utility.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    readFile.mockResolvedValueOnce(template);
+    await Utility.createFileFromTemplate(templatePath, filePath, replacement);
+    expect(writeJson).toHaveBeenCalledWith(filePath, { "@id": modelId }, writeOptions);
+  });
+
+  test("replace all when multiple matched", () => {
+    const template = "{DigitalTwinIdentifier},{DigitalTwinIdentifier}";
```
dilin-MS: >DigitalTwinIdentifier [](start = 23, length = 21)

Same as above.

dilin-MS:
>     if (!name || name.trim() === Constants.EMPTY_STRING) {

Can be replace with 
if (!name.trim())

---
Refers to: src/DigitalTwin/pnp/src/common/utility.ts:58 in 31252c9. [](commit_id = 31252c9c39506d0b60fd96bea5917369b724aed9, deletion_comment = False)

dilin-MS:
>     if (!name || name.trim() === Constants.EMPTY_STRING) {

Same as above.

---
Refers to: src/DigitalTwin/pnp/src/common/utility.ts:77 in 31252c9. [](commit_id = 31252c9c39506d0b60fd96bea5917369b724aed9, deletion_comment = False)

test/digitalTwin/utility.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  });
+
+  test("replace all when not matched", () => {
+    const template = "{digitalTwinIdentifier}";
+    expect(Utility.replaceAll(template, replacement)).toBe(template);
+  });
+
+  test("replace all with multiple pattern", () => {
+    const template = "{DigitalTwinIdentifier},{foo}";
+    expect(Utility.replaceAll(template, replacement)).toBe(modelId + ",bar");
```
dilin-MS: >bar [](start = 71, length = 3)

Use a const variable here.

src/DigitalTwin/pnp/src/common/utility.ts
```diff
@@ -38,12 +38,10 @@ export class Utility {
...
+  static replaceAll(str: string, replacement: Map<string, string>): string {
     const keys = Array.from(replacement.keys());
-    const pattern = new RegExp(keys.join("|"), flag);
+    const pattern = new RegExp(keys.join("|"), "g");
```
chagong: If this module is a general utility I would not suggest to cut it's ability to handle case insensitive condition. Just in case in the future people may want such as a function to do something.

test/digitalTwin/utility.test.ts
chagong: Is this a common programming rule in TS to use a const var wherever a const string/object is needed? I have this question because the programming may look like this:
const str = "hello world";
console.log(str);
instead of:
console.log("hello world");

Isn't that needless? Forgive me that I am new to TS, :)

src/DigitalTwin/pnp/src/common/utility.ts
eriolchan: This function is to replace placeholder for template file. Since the placeholder is defined by ourselves and there is really no need to support case insensitive for placeholder in the future, I remove this logic.

In addition, the current logic doesn't work for case insensitive. Although regular expression can find matched word (digitalTwinIdentifier) with case insensitive, for the replacement, we need to make the keys (DigitalTwinIdentifier) in the replacement map as lower case at first, so that we could find the correct key. The pre-process complicates the code and in most case there is no need for placeholder case insensitive.

src/DigitalTwin/pnp/src/common/utility.ts
eriolchan: This function is to replace placeholder for template file. Since the placeholder is defined by ourselves and there is really no need to support case insensitive for placeholder in the future, I remove this logic.

In addition, the current logic doesn't work for case insensitive. Although regular expression can find matched word (digitalTwinIdentifier) with case insensitive, for the replacement, we need to make the keys (DigitalTwinIdentifier) in the replacement map as lower case at first, so that we could find the correct key. The pre-process complicates the code and in most case there is no need for placeholder case insensitive.

test/digitalTwin/utility.test.ts
eriolchan: Here I want to make template string be more readable/straightforward instead of concatenating one string and one variable.

test/digitalTwin/utility.test.ts
eriolchan: Same as above

test/digitalTwin/utility.test.ts
eriolchan: If possible, I would like to make it as .toBe("urn:model:id:1,bar") to make it more readable/straightforward

eriolchan: fixed

test/digitalTwin/utility.test.ts
eriolchan: fixed

test/digitalTwin/utility.test.ts
eriolchan: fixed

PR #933, opened 2020-02-28
cleanup of sample docs & pics

IvanJobs:
removed some md docs which I'm sure to delete currently.
still work in progress to remove more.


IvanJobs:
Missing concerns about dependency between mds, 
so redesign and close this pr.

PR #934, opened 2020-03-01
Cleanup/sample docs

IvanJobs:
Some redundant markdown docs and pictures are removed.
One bash script was used to find redundant pics with an assumption about the folder structure that all pics under folder media are only used in mds in upper folder. 
The specific logic, please refer to docs/scripts/genImageListToDelete.sh


IvanJobs:
close for wrong branch.

PR #935, opened 2020-03-01
clean up sample docs folder

IvanJobs:
Some redundant markdown docs and pictures are removed.
One bash script was used to find redundant pics with an assumption about the folder structure that all pics under folder media are only used in mds in upper folder.
The specific logic, please refer to docs/scripts/genImageListToDelete.sh


JerryYangKai:
I suggest you to remove devkit-get-started.md and esp32-get-started.md. 
I know some archived docs still need them, we can redirect it to 
https://github.com/Azure-Samples/mxchip-iot-devkit-get-started/blob/master/README.md 
https://github.com/Azure-Samples/esp32-iot-devkit-get-started/blob/master/README.md
or 
https://docs.microsoft.com/en-us/samples/azure-samples/mxchip-iot-devkit-get-started/sample/
https://docs.microsoft.com/en-us/samples/azure-samples/esp32-iot-devkit-get-started/sample/

IvanJobs:
I'm ok with your suggestion.  @JerryYangKai

PR #942, opened 2020-03-09
Support multiple versions for PnP IntelliSense

eriolchan:
- add framework code to support multiple DTDL versions
- apply version rules to diagnostics and auto-completion
- fix bug of diagnostic is not available for the active file when VSCode start
- remove unused parameters for initDigitalTwinCommand

### DIFF
--- a/src/DigitalTwin/pnp/src/intelliSense/digitalTwinCompletionItemProvider.ts
+++ b/src/DigitalTwin/pnp/src/intelliSense/digitalTwinCompletionItemProvider.ts
@@ -104,7 +104,7 @@ export class DigitalTwinCompletionItemProvider implements vscode.CompletionItemP
   }
 
   /**
-   * evaluate if need add separator after offset
+   * evaluate if need to add separator after offset
    * @param text text
    * @param offset offset
    */
--- a/src/DigitalTwin/pnp/src/intelliSense/digitalTwinCompletionItemProvider.ts
+++ b/src/DigitalTwin/pnp/src/intelliSense/digitalTwinCompletionItemProvider.ts
@@ -5,20 +5,29 @@ import * as parser from "jsonc-parser";
 import * as vscode from "vscode";
 import { Constants } from "../common/constants";
 import { DigitalTwinConstants } from "./digitalTwinConstants";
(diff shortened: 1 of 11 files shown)


src/DigitalTwin/pnp/src/intelliSense/digitalTwinCompletionItemProvider.ts
```diff
@@ -97,7 +104,7 @@ export class DigitalTwinCompletionItemProvider implements vscode.CompletionItemP
   }
 
   /**
-   * evaluate the separator after offset
+   * evaluate if need add separator after offset
```
chagong: need to add?

src/extension.ts
```diff
@@ -117,20 +120,24 @@ function initIntelliSense(context: vscode.ExtensionContext): void {
   );
   const diagnosticProvider = new DigitalTwinDiagnosticProvider();
   const activeTextEditor: vscode.TextEditor | undefined = vscode.window.activeTextEditor;
-  if (activeTextEditor) {
-    diagnosticProvider.updateDiagnostics(activeTextEditor.document, diagnosticCollection);
+  if (activeTextEditor && isJsonFile(activeTextEditor.document)) {
```
dooriya: >isJsonFile(activeTextEditor.document)) [](start = 26, length = 38)

Why enable diagnostic if a JSON file is detected rather than a DTDL file?

src/DigitalTwin/pnp/src/intelliSense/digitalTwinCompletionItemProvider.ts
eriolchan: fixed typo

src/extension.ts
eriolchan: Since DTDL file is basically a JSON file, narrow down the file type to reduce unnecessary calls on other file types. For Auto-compeletion and Hover, we also use document selector for this purpose.

src/extension.ts
```diff
@@ -93,7 +93,10 @@ function initCommand(
   context.subscriptions.push(vscode.commands.registerCommand(command, callback));
 }
 
-// DigitalTwin extension part
+function isJsonFile(document: vscode.TextDocument): boolean {
```
chagong: Looks like this function is only called by initIntelliSense(). So can we put it within initIntelliSense()?

PR #944, opened 2020-03-10
feat: add DTDL v2 definition change

eriolchan:
- Interface Instance is renamed to Component.
- context value of v2 is "http://azure.com/v2/contexts/Model.json"

### DIFF
--- a/src/DigitalTwin/pnp/src/intelliSense/digitalTwinGraph.ts
+++ b/src/DigitalTwin/pnp/src/intelliSense/digitalTwinGraph.ts
@@ -250,7 +250,6 @@ export class DigitalTwinGraph {
    * @param name name
    */
   private getId(name: string): string {
-    // TODO(erichen): check if it works for v2 definition
     return this.vocabulary + name;
   }
 
@@ -268,7 +267,6 @@ export class DigitalTwinGraph {
         continue;
       }
       const value = context[key];
-      // TODO(erichen): check if need internal id for multiple versions
       if (typeof value === "string") {
         id = this.getId(value);
         this.contextNodes.set(id, { name: key, container: ContainerType.None });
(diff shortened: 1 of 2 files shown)


src/DigitalTwin/pnp/src/intelliSense/digitalTwinGraph.ts
```diff
@@ -579,7 +577,89 @@ export class DigitalTwinGraph {
    * patch definition change for different DTDL version
    */
   private handleDefinitionChange(): void {
-    // DTDL v2
+    // DTDL v2 change
+    const version = 2;
+    // rename InterfaceInstance to Component
+    const componentNode: ClassNode | undefined = this.renameClassNode(
+      DigitalTwinConstants.INTERFACE_INSTANCE_NODE,
+      DigitalTwinConstants.COMPONENT_NODE,
```
dooriya: Why not re-build the graph? Then you don't need to do such replacement.

eriolchan: The major reason is that we have no graph to build so far for DTDL v2 definition, and I doubt if the build tool still works for the new JSON-LD files with SHACL syntax. Since there are so many uncertain factors about the definition file, I prefer not directly modifing graph.json (which is error-prone), but use code to adjust DigitalTwin graph instead. This could unblock our progress on supporing PnP Refresh. Once we have final and stable graph, we coud use it and replace this code.

For multiple version support, my solution is to provide one DigitalTwin graph for all versions, and search available definition by version for IntelliSense. The reason I don't choose maintain one graph by each version is that.
- Loading graph is a time-consuming operation, if user modify the context version of file or switch to another file with different version, reloading graph will make it slow to user experience.
- The definition change of v2 or later is based on v1, and are relatively small changes compared with the whole v1 definitions. Holding multiple graph versions with most duplicated definitions in memory is not worth.

dooriya: I see, but it would become a great concern to "use code to adjust DigitalTwin graph instead", since the language definition can vary greatly between multiple version, not just adding/removing some node, but also change the node definition. This may be very complicated and increase our maintenance costs. That's why I suggest using different graph for different language version. But it would be good to share the same graph if the language definition doesn't change too much in DTDL v2.

As you mentioned above, loading graph is a time-consuming operation, how about using some cache mechanism or pre-load the graph in advance?

eriolchan: Yes, it is the trade-off between code complexity and app resource (both time and memory). It is a short-term solution before PnP GA. Once DTDL definition is final and stable with GA, we could revisit here to make a better solution for long-term.

PR #947, opened 2020-03-18
feat: send PnP usage telemetry when opening model file

eriolchan:


src/extension.ts
```diff
@@ -150,11 +146,25 @@ function initIntelliSense(context: vscode.ExtensionContext): void {
...
   );
+  // send DigitalTwin usage telemetry
+  context.subscriptions.push(
+    vscode.workspace.onDidOpenTextDocument(document => {
+      if (IntelliSenseUtility.isJsonFile(document)) {
+        const modelContent: ModelContent | undefined = IntelliSenseUtility.parseDigitalTwinModel(document.getText());
+        if (modelContent) {
+          const telemetryContext: TelemetryContext = telemetryWorker.createContext();
+          telemetryContext.properties.result = TelemetryResult.Succeeded;
+          telemetryContext.properties.version = modelContent.version.toString();
```
blackchoey: can `version` be something like `DTDLVersion`? This could help people quickly understand what does this property mean with its name.

eriolchan: fixed. Change to dtdlVerson

PR #949, opened 2020-03-19
Fix config dps and uds

JerryYangKai:


dilin-MS:
@JerryYangKai commit message can be more specific like "fix bug: unable to config dps and uds" for better tracking in the future.

PR #952, opened 2020-03-24
add changelog for 0.13.0

dilin-MS:


src/Models/AZ3166Device.ts
```diff
@@ -762,11 +762,12 @@ export class AZ3166Device extends ArduinoDeviceBase {
   }
 
   private async getComList(): Promise<SerialPortInfo[]> {
-    const lists = await AZ3166Device.serialport.list();
-    if (!lists.length) {
-      vscode.window.showInformationMessage("No serial port is available.");
+    try {
```
eriolchan: If we do nothing with err, can we remove the wrapper of try...catch?

dilin-MS: fixed.

PR #957, opened 2020-03-26
webpack bundle improvement

chagong:
- Improve webpack bundle method, try as much as possible to bundle .js files into one dist/extension.js;
- For those external modules are not webpack friendly, leave it under node_modules/ and make sure pack them into vsix;
- Remove unnecessary files in .vscodeignore;
- Significantly reduce the size of the extension;

[AB#1692801](https://dev.azure.com/mseng/c92b8a86-6e9e-4f54-ae64-7b6a5e479667/_workitems/edit/1692801)
[AB#1692822](https://dev.azure.com/mseng/c92b8a86-6e9e-4f54-ae64-7b6a5e479667/_workitems/edit/1692822)
[AB#1696796](https://dev.azure.com/mseng/c92b8a86-6e9e-4f54-ae64-7b6a5e479667/_workitems/edit/1696796)


.vscodeignore
```diff
@@ -1,17 +1,45 @@
...
+!node_modules/@serialport/bindings
+!node_modules/@serialport/binding-abstract
+!node_modules/@serialport/parser-readline
+!node_modules/@serialport/parser-delimiter
+!node_modules/debug
+!node_modules/ms
+node_modules
+out
+src
+vendor
```
eriolchan, line 28: Have we removed vendor folder with the change of usb-detection?

.vscodeignore
```diff
@@ -1,17 +1,45 @@
...
+!node_modules/errlop
+!node_modules/extract-opts
+!node_modules/typechecker
+!node_modules/eachr
+!node_modules/getos
+!node_modules/node-usb-native/native
+!node_modules/usb-detection/build/Release/detection.node
+node_modules/usb-detection/doc
+node_modules/usb-detection/test
+!node_modules/usb-detection/*.*
```
eriolchan, line 15: Is it valid to use "!node_modules/usb-detection" instead of "!node_modules/usb-detection/*.*" here?

.vscodeignore
```diff
@@ -1,17 +1,45 @@
...
+!node_modules/@serialport/binding-abstract
+!node_modules/@serialport/parser-readline
+!node_modules/@serialport/parser-delimiter
+!node_modules/debug
+!node_modules/ms
+node_modules
+out
+src
+vendor
 gif/**
```
eriolchan, line 29: Can we align the glob of folder with others (e.g. src, out), and remove "**"?

.vscodeignore
```diff
@@ -1,17 +1,45 @@
...
-scripts/**
\ No newline at end of file
+scripts/**
+__mocks__/**
+azure-pipelines.yml
+coverage/
+devicedevex_dockerfiles
+jest.config.js
+updateDynamicResourceLink.js
+**/*.map
```
eriolchan, line 45: May be also add eslint related files, such as .eslintignore. Since we don't use Travis for CI, we could remove travis.yml in the source code.

src/DigitalTwin/CodeGeneratorCore.ts
```diff
@@ -533,7 +533,8 @@ export class CodeGeneratorCore {
     context: vscode.ExtensionContext,
     channel: vscode.OutputChannel
   ): Promise<CodeGeneratorConfigItem | null> {
-    const extensionPackage = require(context.asAbsolutePath("./package.json"));
+    //const extensionPackage = require(context.asAbsolutePath("./package.json"));
```
eriolchan, line 536: It is better to remove commented code

src/DigitalTwin/CodeGeneratorCore.ts
```diff
@@ -533,7 +533,8 @@ export class CodeGeneratorCore {
     context: vscode.ExtensionContext,
     channel: vscode.OutputChannel
   ): Promise<CodeGeneratorConfigItem | null> {
-    const extensionPackage = require(context.asAbsolutePath("./package.json"));
+    //const extensionPackage = require(context.asAbsolutePath("./package.json"));
+    const extensionPackage = JSON.parse(fs.readFileSync(context.asAbsolutePath("./package.json"), "utf8"));
```
eriolchan, line 537: I think you could use getJsonContent to get JSON content, which is in src\DigitalTwin\pnp\src\common\utility.ts

src/boardProvider.ts
```diff
@@ -4,6 +4,7 @@
 import { FileNames } from "./constants";
 import { Board } from "./Models/Interfaces/Board";
 import * as path from "path";
+import * as fs from "fs-plus";
```
eriolchan, line 7: I recommend to use fs-extra module (which is already in package.json) instead of fs-plus

src/exampleExplorer.ts
```diff
@@ -273,7 +273,8 @@ export class ExampleExplorer {
     const boardList = context.asAbsolutePath(
       path.join(FileNames.resourcesFolderName, FileNames.templatesFolderName, FileNames.boardListFileName)
     );
-    const boardsJson: { boards: Board[] } = require(boardList);
+    //const boardsJson: { boards: Board[] } = require(boardList);
```
eriolchan, line 276: Can you help check with Kai/Di if we still need example Explorer in Workbench? I remember Kai has removed the command palette of example

.vscodeignore
chagong: vendor folder has already been deleted by dilin's change to node-usb-native module.

.vscodeignore
chagong: There are a sub-dir in usb-detection contains some cpp files, which I don't wanna pack in, but I need to pack the source files under usb-detection. So wrote like this, but I think probably i can remove the previous 2 lines. I will change it.

.vscodeignore
chagong: Sure.

.vscodeignore
chagong: Sure.

src/DigitalTwin/CodeGeneratorCore.ts
chagong: OK, I just forgot to do so.

src/DigitalTwin/CodeGeneratorCore.ts
chagong: OK

src/boardProvider.ts
chagong: Sure.

src/exampleExplorer.ts
chagong: I think we still need the example explorer, we have a command for opening sample page and let user download sample code. @JerryYangKai @dilin-MS can you confirm?

.vscodeignore
```diff
@@ -1,17 +1,44 @@
...
+!node_modules/file-uri-to-path
+!node_modules/@serialport/bindings
+!node_modules/@serialport/binding-abstract
+!node_modules/@serialport/parser-readline
+!node_modules/@serialport/parser-delimiter
+!node_modules/debug
+!node_modules/ms
+node_modules
+out
+src
```
dilin-MS, line 25: Are we still have `src/` folder?

.vscodeignore
```diff
@@ -1,17 +1,44 @@
...
-out/**/*.map
-out/**/*.ts
-src/**
-gif/**
-hooks/**
-test/**
-docs/**
+.vscode
+.vscode-test
+!node_modules/keytar
```
dilin-MS, line 3: Can you add a comment for these vscode ignore files/folders? For example, some are for customized webpack, some are just for CI/CD. This can help developers understand when they should add files in this .vscodeignore file.

.vscodeignore
chagong: No, this line actually exclude src folder. The old way is to retain the whole out/src folder, now we don't keep them all instead of one single .js under dist/ folder.

.vscodeignore
dilin-MS: Then maybe we should remove the `vendor` folder here?

.vscodeignore
chagong: OK.

.vscodeignore
```diff
@@ -1,17 +1,44 @@
...
 gts.cmd
-scripts/**
\ No newline at end of file
+scripts
+__mocks__
+azure-pipelines.yml
+coverage
+devicedevex_dockerfiles
+jest.config.js
+updateDynamicResourceLink.js
```
dilin-MS, line 43: Could you help check if this file is useless now. If so, we can remove the file from source code.

src/Models/AZ3166Device.ts
```diff
@@ -35,9 +35,9 @@ import { SystemResourceNotFoundError } from "../common/Error/SystemErrors/System
 import { AzureConfigFileHandler } from "./AzureComponentConfig";
 import { ComponentType } from "./Interfaces/Component";
 
-const impor = require("impor")(__dirname);
-const forEach = impor("lodash.foreach") as typeof import("lodash.foreach");
-const trimStart = impor("lodash.trimstart") as typeof import("lodash.trimstart");
+const importLazy = require("import-lazy");
+const forEach = importLazy(() => require("lodash.foreach"))() as typeof import("lodash.foreach");
```
dilin-MS, line 39: Why do we need to have `as typeof import("lodash.foreach")` here? It looks different from other import-lazy requirement statements.

.vscodeignore
chagong: Sure.

src/exampleExplorer.ts
dilin-MS: We still have the example Explorer. What you mention may be the `Azure IoT Device Workbench: Help` command, which will open a welcome page showing a oil pumping machine in the past and the page was later deleted by @JerryYangKai . Now the command will lead to a external URL.

src/Models/AZ3166Device.ts
chagong: This is to avoid warning of using implicitly any type.

